PR #91, opened 2018-03-15
Needed changes to eliminiate emory-dev

birm:
emory-dev has been a bit of a mess, so I created this branch to try to just add the needed changes.

It's worth noting that some changes seem to mess with OSDImageMetaData; as far as I can tell this is an issue on develop as well, though.

@tdiprima - Does this look like it breaks anything?
@lastlegion - Was there anything else on emory-dev that I missed here?


tdiprima:
@birm a change was made to shared/osdHeader.php, the javascript src was changed from `<script src="/js/config.js"></script>` to `<script src="../js/config.js"></script>`.  That's actually wrong.  Please change it back.

birm:
I've reverted that change. The reason I wanted a relative path here is for our reverse proxy deployments; for example, in https://wolf.cci.emory.edu/camic-anno/, it will look for https://wolf.cci.emory.edu/js/config.js instead of https://wolf.cci.emory.edu/camic-anno/js/config.js

tdiprima:
It also means we have to test.  I'm testing this, and I'm getting:
```
Uncaught TypeError: Cannot set property 'style' of null
    at spyglass_init_camic.js:41
```

tdiprima:
And
```
DOMException: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
    at StateManager.decode 
```

PR #96, opened 2018-03-20
Autotesting

birm:
A basic test. Please don't merge until we know travis works for this.
@lastlegion can you turn on travis for this repo?


birm:
There are some oddities on the travis side that I can't diagnose, so I had to discard an assertion that worked locally. So, this is a not-very comprehensive test set for this repo, but should allow for more tests to be written in the future.

PR #101, opened 2018-03-27
Human markups [WIP]

birm:
changes:

- Use getMultipleAnnots instead of getMultipleAnnotationsNew
- Add rectangle and pencil buttons
- Use the user's username as execution_id
- Post to getAnnotSpatial_sc in addition to getAnnotSpatial to register the algorithm for the selector


birm:
With a switch to getMultipleAnnots, pencil and rect both work, and so I've added them back.

birm:
the post method of getAnnotSpatial_sc registers the user to show up in the algorithm selector, while getAnnotSpatial does not do this. So, I don't think any further adjustment is needed here, until we do api harmonization.

tdiprima:
Thanks, It looked like the addition of the button(s) was commented out.  Checked, looks ok.
`tool.append(this.rectbutton)`

PR #108, opened 2018-04-05
Start for unified api endpoints

birm:
Conceptually, it makes sense for each type of document to have its own url which works for all applicable http verbs. this is my start of trying to get this to work.


birm:
part of https://github.com/SBU-BMI/quip_distro/issues/200

birm:
In the process of testing now; I'll update with results

PR #111, opened 2018-04-13
revert sidesplit and selections as class obj

birm:
I don't know how to test your applications, but this should hopefully undo the changes which were part of sidesplit.


tdiprima:
Ryan, `osdAnnotationTools.js` needs fixing as well.

tdiprima:
ok... DisplaySelector.js... `annotools1.SELECTED_ALGORITHM_LIST = [e.target.value];`... according to Ryan is part of sidesplit and it shouldn't be included anymore.

Team will test to make sure the app is functioning.

PR #113, opened 2018-04-16
Sidesplit without use of Annotools Directly

birm:


birm:
I've asked for a svg icon for this tool, so that may be added either before/after merge.
This should not impact other existing functionality, and it is only included in osdCamicroscope.php. That said, I'm happy to better testing on other components if they're identified.

tdiprima:
> I've asked for a svg icon for this tool...
Can we help with that?

PR #115, opened 2018-04-26
Mocha functional and unit test examples and setup

birm:


birm:
Good catch on 1. For 2, it's run with this command: https://github.com/camicroscope/caMicroscope/blob/cb0338cf6f1e087e75a000ac060933cb88d06587/.travis.yml#L20

PR #122, opened 2018-05-04
Bug 218

tdiprima:
**Issue 1:** Roll over ROI does not change cursor to hand.
**Main cause of issue:** was the addition of style "pointer-events: none;".

Removed.  (1) Problem solved.


tdiprima:
Issue SBU-BMI/quip_distro#218

PR #133, opened 2018-11-08
relative links

tdiprima:
Hi guys,

I made a change to:
```
core/CaMic.js 
core/Store.js 
js/init.js
```

Changed `base=./data/` to `base=/data/`,
`./img/Slide/` to `/img/Slide/`,
and
`./table.html` to `/table.html`,
because if you want to do use those functions in another directory, you couldn't, because it uses the relative link.  Example: `/test/extension/draw-polygon-test.html`

Is this OK?

### DIFF
--- a/core/CaMic.js
+++ b/core/CaMic.js
@@ -6,13 +6,13 @@
 
 class CaMic{
   /**
-  * 
+  *
   * create a camic core instance
   * @param divId - the div id to inject openseadragon into
   * @param slideId - the id of the slide to load
   * @property slideId - the slide id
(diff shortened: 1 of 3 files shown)


birm:
Hmm, what does this do when we have a reverse proxy using a suffix, like http://wolf.cci.emory.edu/camic_uat/ ?

tdiprima:
@birm great question!  By any chance, would you be able to test branch `temp1`?  Or lmk how to set up like yours?

birm:
Per Tammy, this was resolved in a way that didn't cause issue with our sometimes strange deployment strategy ;)

PR #168, opened 2019-04-20
Added support for uploading multiple files at once

pranavbudhwant:


birm:
Great start! Some of the functionality interferes, though, such as checking info and thumbnails. Also, it's misleading to just show post count as 1 for multiple. To get around this, you could post both together, or just change the output to differ from the data service's literal output.

pranavbudhwant:
@birm Sir, the POST does actually show the count of all the images uploaded, but sometimes it shows less, same with the CHECK functionality. It shows the correct data after clicking on the buttons more than once. I'm not sure why there's this problem though. 
Here's what I can see on my system:
Screenshot from 2019-04-21 15-50-17
Screenshot from 2019-04-21 15-50-37
Screenshot from 2019-04-21 15-50-52

PR #202, opened 2019-09-09
hotfix

tdiprima:
Deleted line 23 of pathdb_package, because it errors there instead of checking the variable first (line 39, it checks it then utilizes it).

`dist` needs to be rebuilt.  I'm getting an error on `parcel build package/pathdb_package.js`  Can you please assist?


birm:
Yes, the package does this twice for some reason; merging to develop, rebuilding, and promoting through.

PR #222, opened 2020-03-01
Fixed some spellings

AniketDogra:
Fixed some spelling mistakes which I encountered while browsing the code.


AniketDogra:
@birm Please have a look

PR #223, opened 2020-03-02
Feature/remove model

chinzhiweiblank:
Add a new column Remove Models.
Underneath is a button that removes the model from the model store and key from the database.
deleteModel1
deleteModel2
deleteModel3


Insiyaa:
Good work! I noticed that after clicking and confirming to remove the model, the entry is not removed from the table, instead the table has to be closed and opened again to see the changes. Can you please look into that?

chinzhiweiblank:
@Insiyaa Rectified the issue mentioned.

Insiyaa:
Thank you, it works as required!

PR #225, opened 2020-03-03
adding dependency check feature

darkCavalier11:
**what it does?**
**depcheck** is a npm module used to keep track of the dependency, dev-dependencies and point out unused,  and where the dependencies are used.

**changes made**
- Installed depcheck at package.json
- Added a new file `depcheck.js` in main directory

**screenshot**
depcheck
 
PS: if u are getting error like **vue version mismatch** then update vue and vue-template-compiler to same version.


birm:
You're right in that we haven't updated package.json to remove old packages. Though, I'm kind of on the fence on adding this dependency. I guess it could be a second CI check, but we don't often add or remove our dependencies. Is this still a useful thing to have?

birm:
Closing since the container isn't using node anymore.

PR #231, opened 2020-03-10
Fixes Memory Leak

### LINKED ISSUE #228: Additional Memory being allocated every time Prediction is done
When the predict feature is used on the same slide using same model , each time a new area is selected and Prediction is done new memory(tensors) is allocated . Thus after a few number of predictions the memory is maxed out . 

Selection_003

@Insiyaa @birm  How do we measure the memory consumption of the model ? ( To add it to  ``` model info ``` table for issue #215 as with every trial the memory consumption increases )

leoarc:
Fixes #228 
Before :

Selection_004

After :

Selection_006

All unused tensors  inside the ``` runPredict() ``` are being disposed . Performance for multiple predictions has improved a lot. Running multiple predictions dosen't overflow memory now.


leoarc:
@Insiyaa  @birm Please have a look at it and suggest necessary changes,

apps/model/model.js
```diff
@@ -508,43 +512,56 @@ function runPredict(key) {
         fullResCvs.width = l_img.width;
         fullResCvs.getContext('2d').drawImage(l_img, 0, 0);
 
-        let imgData = fullResCvs.getContext('2d').getImageData(0,0,fullResCvs.width,fullResCvs.height);
 
-        const img = tf.browser.fromPixels(imgData).toFloat();
+        let imgData = fullResCvs.getContext('2d').getImageData(0,0,fullResCvs.width,fullResCvs.height);
         let img2;
+       tf.tidy(()=>{
```
Insiyaa: img2 is hanging after tf.tidy

apps/model/model.js
```diff
@@ -508,43 +512,56 @@ function runPredict(key) {
...
           normalized = img2.sub(mean);
+          mean.dispose();
           // normalized.mean().print(true); // Uncomment to check mean value.
           // let min = img2.min();
           // let max = img2.max();
           // let normalized = img2.sub(min).div(max.sub(min));
         } else {
           // Pixel Standardization: scale pixel values to have a zero mean and unit variance.
-         
+            tf.tidy(()=> {
```
Insiyaa: Where is `normalized` returned to

Insiyaa:
The [mobilenet example](https://github.com/tensorflow/tfjs-examples/blob/908ee32750ba750a14d15caeb53115e2d3dda2b3/mobilenet/index.js#L61) might help you with placing these calls.

apps/model/model.js
leoarc: to keep ``` tf.tidy() ``` from destroying the tensor it needs to be returned. ``` normalized ``` is being used afterwards so it doesn't need to be destroyed.

Insiyaa: The returned object has to be assigned

leoarc: I am using ``` tf.keep() ``` instead of return. Does this solve the issue?

leoarc:
@Insiyaa  Tried fixing the issues . Let me know if they are ok .

apps/model/model.js
Insiyaa: It still needs to be assigned, please refer the usage here: https://js.tensorflow.org/api/latest/#keep

Insiyaa:
You can enclose into `tf.tidy` starting from [here](https://github.com/camicroscope/caMicroscope/blob/c219afb62cbdb970d19d1df948bcaa461fc780de/apps/model/model.js#L513) till the end of the for block. On the `predict` call, remove `await` and make it `dataSync()`. This should help, the same changes are to be done for segment app.

leoarc:
> On the predict call, remove await and make it dataSync(). 



This doesn't work . As given in the documentation [here](https://js.tensorflow.org/api/0.6.1/#tf.Tensor.dataSync) ``` tf.dataSync() ``` is used to download values from a tensor (which is used to store tensor to a js array [link](https://meowni.ca/posts/on-tfjs-datasync/) ) . I don't know what it is doing [here](https://github.com/camicroscope/caMicroscope/blob/17552ad812932e7fdbe13e0a2f8c3ddc5113d483/apps/model/model.js#L475) but code runs fine if it is commented. So removing ``` await ``` in the predict call makes it return a promise to ``` values ``` and thus causes errors in the next steps.

leoarc:
@Insiyaa As you mentioned earlier , now I have assigned all the variables returned by ``` tf.tidy() ```. 
Please review.

Insiyaa:
Oh yes, that is just a warmup run and we don't really need to get the results, so removing that line would do just good.
What I mean to suggest is that just one `tf.tidy` call enclosing everything is needed,  [this line](https://github.com/camicroscope/caMicroscope/blob/c219afb62cbdb970d19d1df948bcaa461fc780de/apps/model/model.js#L548) being included or not depends on whether the values are obtained sync (_i.e. -_ `let values =  model.predict(batched).dataSync();` _goes inside the tidy call_) or async (_let it stay outside_). Since the predictions happen patch-wise, I believe it should go about synchronously.

leoarc:
@Insiyaa So i started from the beginning. Now the for loop has only call to ``` tf.tidy() ``` . There is one other call to ``` tf.tidy() ```  [here](https://github.com/camicroscope/caMicroscope/blob/17552ad812932e7fdbe13e0a2f8c3ddc5113d483/apps/model/model.js#L473) used to ``` dispose ``` the ``` tf.zeros(..) ``` tensor.  Please review .

apps/model/model.js
```diff
@@ -471,10 +473,11 @@ function runPredict(key) {
     self.showProgress("Model loaded...");
 
     // Warmup the model before using real data.
+    tf.tidy(()=>{
     const warmupResult = model.predict(tf.zeros([1, image_size, image_size, input_channels]));
     warmupResult.dataSync();
```
Insiyaa: I realise this isn't required, could you remove it?

Insiyaa:
That seems good, thank you! The same additions would go into segment.js. Also, could you squash the commits into one with a meaningful commit message :)

birm:
If you prefer not to mess with rebasing, we can squash when merging the PR itself.

leoarc:
So i am facing a problem. This [line](https://github.com/camicroscope/caMicroscope/blob/17552ad812932e7fdbe13e0a2f8c3ddc5113d483/apps/segment/segment.js#L768) is returning a tensor which I am not being able to dispose . The documentation says ``` tf.browser.toPixels  ```  returns a promise . So  I am not able to do ``` let x = await tf.toPixels(...) ``` and ``` x.dispose() ``` .  And I suppose this function has to be asynchronous .Any ideas ? @Insiyaa

leoarc:
@birm  @Insiyaa Should we close this for now ? It still has the issue which I mentioned in the last comment . The issue has been mentioned [here](https://github.com/tensorflow/tfjs/issues/264) and though it is showing solved I think it isn't . Anyways it must be a internal leak . Please review.

Insiyaa:
In my view, all the required additions are done. You're right; there is probably some leak inside `tf.browser.toPixels`, we can keep track of it. Also, you might want to look into backends if it helps with this issue. 
@birm would you like to suggest any changes?

PR #226, opened 2020-03-04
combined load slide page and table page

vikasgola:
## Added Load slide option to table page and removed load slide page.

image

image

PR #232, opened 2020-03-10
impoving the instructions in #226

viditvarshney:
### **Uploading instructions should be changed and improved.**

issue


viditvarshney:
@birm  the changes in the instruction to look better than previously.

PR #233, opened 2020-03-10
Improved mismatch of Navbars, added favicon, remove unneccessory code

viditvarshney:
- When I was working on some other problem I noticed the mismatch of the navbars as shown below.

CaMicroscope-Data-Table
- Added favicon 
- remove the unnecessary css


viditvarshney:
@birm @nanli-emory  please have a look :+1:

PR #234, opened 2020-03-11
fixing the testing functionality of test directory

darkCavalier11:
**what this PR does?**
fixed the test/extension directory that used for testing the code.

**changes made**
i found there is a lot of error in the `test/extension/draw-polygon-test.html` file, like wrong syntax in importing the css and js file location, module not found error etc . when i opened the html file i found the following page load up with errors.
Screenshot from 2020-03-11 17-09-36
so i make the test a much more friendly and structured with the help of webpack-dev-server. a number of dev-dependencies are installed along with webpack loader to support the process.
made separated file for all css and js imports and bundle it dynamically to `dist` directory and serving it at `localhost:8080` by webpack as default. so here is the result----

**screenshot now**
Screenshot from 2020-03-11 17-14-17

 **follow ups**
their is still some error in css file loading that i will fix soon.


darkCavalier11:
@birm plz review

PR #237, opened 2020-03-15
added webpack functionality to test/extension

darkCavalier11:
**what this PR does?**
fixed the test/extension directory that used for testing the code.

**changes made**
i found there is a lot of error in the test/extension/draw-polygon-test.html file, like wrong syntax in importing the css and js file location, module not found error etc . when i opened the html file i found the following page load up with errors.
before

so i make the test a much more friendly and structured with the help of webpack-dev-server. a number of dev-dependencies are installed along with webpack loader to support the process.
made separated file for all css and js imports and bundle it dynamically to dist directory and serving it at localhost:8080 by webpack as default. so here is the result----

screenshot now
after

**follow ups**
their is still some error in css file loading that i will fix soon.


darkCavalier11:
the test is failing because this is a client side code with web platform APIs which meant to run on browser instead it is running on node which show many reference errors when stumble across the syntax like `import` or `window` etc. @birm can u plz suggest something?

birm:
I'm still not sure exactly what you're trying to do. Is something wrong or broken about the current test system? Or is this just about readability?
I've seen references to a node module called [esm](https://www.npmjs.com/package/esm) which _may_ resolve your issue, but may not. Good luck!

darkCavalier11:
i was going through the `test/extension/draw-polygon-test.html` file. i found there are multiple errors in location syntax while importing css and js files. i fix all of them. but there still some errors showing in the browser because the js files being loaded are not in sequence as they should be. i thought best to bundle all css and js files together in a single file at `dist` directory so that the html file link to it and webpack being served at local sever. everything worked fine, but when i do `npm run test` it failed because the bundled file being run by node. obviously node will fail because it is a bundle of css and client side js files hence the travis ci failing.

birm:
Does https://github.com/camicroscope/caMicroscope/commit/83cb9ae78a4abd7d92245663623a1a5ced948908 resolve this issue too?

darkCavalier11:
Yes

PR #246, opened 2020-03-24
Added title to button

amritansh22:
Added title to close button in segment panel. Now when we hover to close button, we would see a text displayed "Close Panel"


birm:
@Insiyaa does this look good to you?

Insiyaa:
Yes

PR #247, opened 2020-03-25
Bug Fix

AbdulRashidReshamwala:
There was a missing id attribute in the input tag which resulted in failing to add a event listner
Before
Screenshot from 2020-03-25 19-54-15
After
Screenshot from 2020-03-25 19-59-43


birm:
Closing in favor of #249  since new commits changed the scope.

PR #249, opened 2020-03-25
Show Upload Options on Table when no slides

### LINKED ISSUE #248: Missing Ability to upload files
**Bug**
commit 39d2d683dc73619190b3f8b57d8f69272dbfcb63 seems to have removed apps/loader/loader.html which was the primary way to upload files is there another way to upload files or was it removed by mistake
**To Reproduce**
 Going to 'http://localhost:4010/apps/loader/loader.html' errors out.

**Expected behavior**
Show the uploading UI

**Screenshots**
Screenshot from 2020-03-25 20-31-40

AbdulRashidReshamwala:
Due to line 286 in apps/table.html both the upload and reload buttons as well as the upload modal was removed 
**Before**
Screenshot from 2020-03-25 20-52-40
**After**
Screenshot from 2020-03-25 20-51-52
Screenshot from 2020-03-25 20-58-55

fixes #248


birm:
For future reference, it's easier to review PRs when you use a different topic branch for each change.

AbdulRashidReshamwala:
@viditvarshney  take a look at attached screenshots earlier when there was no data the modal and the buttons to upload were being overwritten with "No Data Found  x..x"

viditvarshney:
no, i am asking of this particular commit which has a message of "fix upload issues"

viditvarshney:
sir when the files are uploaded then the upload and reload buttons are automatically above the table 
76673615-b5707000-65cc-11ea-983d-87f6944bc95c

birm:
So, what has happened is that when no slides are present, the "no data" message would overwrite the table. That also meant that the upload/reload buttons were removed from the DOM. I missed this since I usually test from the test_seed data which adds a slide by default.
@AbdulRashidReshamwala has fixed this, by moving it below the table. I just want them to move it back above :)

AbdulRashidReshamwala:
I have made the required changes.

viditvarshney:
ohh yeah sir i recently noticed that 🙂

PR #253, opened 2020-03-26
Custom colors for heatmap 

rahuldeepattri:
Added options to select color for heat maps in both gradient and binal mode. <br>
image

image

closes https://github.com/camicroscope/Distro/issues/93


components/heatmapcontrol/heatmapcontrol.js
```diff
@@ -240,3 +316,33 @@ function createOpacities(container, field, changeFunc){
...
+		container.appendChild(div);
+	}
+}
+// returns selected colors for intervals
+function getColors(container){
+	const rs = [];
+	$(container).children().each(function (index,colorDiv) {
+		rs.push(colorDiv.querySelector('input').value)
+	});
+	return rs.reverse();
```
birm: I'll invite @nanli-emory to comment on this too, but I think it makes more sense to show the lower value heatmap bins first.
```suggestion
	return rs;
```

rahuldeepattri: Agreed, i used the reverse order because the code in legend was in that order. I have a feeling that will need to update legend icons code also. I'll check and update accordingly.

rahuldeepattri: Also, @birm @nanli-emory what should be maximum limit on number of intervals. As of now i have made min 2 and max 5 intervals.

rahuldeepattri: New order
image

nanli-emory: max 10 intervals

rahuldeepattri: The logic for gradient should go from lighter colors for less values to more darker colors for higher values, or from blue to red? 
image

Above screen shot has gradient from white to blue for lower to higher values. Of course user can change color of each interval. If this isn't appropriate please suggest the 10 colors for which gradient should be formed.

components/heatmapcontrol/heatmapcontrol.js
```diff
@@ -86,20 +102,52 @@ HeatmapControl.prototype.__refresh = function(){
...
+	//min max logic for valid number of intervals
+	$(colorsLegendPanel).find('#legendIntervals').on('change', function(e){
+		var min=parseFloat($(this).attr('min'));
+		var max=parseFloat($(this).attr('max'));
+		var curr=parseFloat($(this).val());
+		if (curr > max) { $(this).val(max); var changed=true; }
+		if (curr < min) { $(this).val(min); var changed=true; }
+		if (changed) {
+		  $warning = $(colorsLegendPanel).find('.warning')
+		  $warning.text('Only values in' + min + ' through ' + max + ' allowed.');
```
birm:
```suggestion
		  $warning.text('Only values in ' + min + ' through ' + max + ' allowed.');
```

components/heatmapcontrol/heatmapcontrol.js
rahuldeepattri: waiting for your reply @nanli-emory.

PR #260, opened 2020-03-27
selection issue fix

leoarc:
If the region selected for prediction is smaller than the input size requirement of the model , the page got stuck at :

Selection_027

And there's a corresponding memory leak : 

Selection_026

Fixed this and added an alert to tell user to select a larger section . But I cannot understand why  ```  self.close( )  ``` is not working for closing the ``` modelpanel ```

Please review @Insiyaa  @birm


Insiyaa:
Thanks for fixing this redundant computation!
`self.close()` is working, due to [this](https://github.com/leoarc/caMicroscope/blob/a2445b0d70bf5df3023f6f7eac28d335c305096f/apps/model/model.js#L379) line it remains open. A temporary fix maybe to put some delay before `self.close()` :p.
 @birm might have some better suggestions.

birm:
ModelPanel.prototype.open is an async function, so it returns a promise, so you could use .then to call something after it's finished.

leoarc:
@birm @Insiyaa I know this is not a clean way but does this look good to you ? ( Also sorry for so many commits , I messed up trying to resolve conflicts)

birm:
We'll squash on merge :)

PR #263, opened 2020-03-28
Feature to change class list

amritansh22:
Earlier, class list once made can't be altered. This caused the following problems:

- There may be some spelling errors needed to be fixed.
- We may need to add some more classes
- We may need to remove some classes

So I added this functionality.

The new info modal box looks like:
Screenshot (144)

Then a modal box is displayed to take input 

Screenshot (145)

Then a message is displayed upon success. 
Screenshot (146)
And the changes are saved

### DIFF
--- a/apps/model/model.html
+++ b/apps/model/model.html
@@ -185,7 +185,7 @@
     }
 
     #mtable thead th:nth-child(2) {
-      width: 30%;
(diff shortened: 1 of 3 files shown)


birm:
I'll give this a test and ping @Insiyaa to do the same.

PR #264, opened 2020-03-29
Prevent popuppanel overflow

mgautam98:
The notes oerflows and gos out of the popuppanel. This will prevent it from overflow. I have inserted new lines when they go out of popuppanel.
This  used to happen 
Screenshot from 2020-03-29 05-34-16

Now it will not go out of the popuppanel
Screenshot from 2020-03-29 05-32-24


birm:
Thanks for the fix!

PR #265, opened 2020-03-29
prevent default for shortkeys

mgautam98:
The shortkeys work fine for nanoborb. But for the browsers it doesn't work. The browser shortcut keys overrides our shortcuts. 

e.g. ctrl + r is for refresh in chrome. So Instead of measurement the page refreshes.


mgautam98:
Okay. I will make the changes.

mgautam98:
@birm made the requested changes

PR #275, opened 2020-03-29
fix spyglass out of window

mgautam98:
The spyglass goes out of the window at all of the corners like this

spyglass

This this fix it won't go out of window at the corners

Screenshot from 2020-03-30 04-00-37
Screenshot from 2020-03-30 04-01-33


birm:
Good, the CI linting is working. Sorry to have introduced a style guide in the middle of your PR work.

You can probably auto-fix these with `eslint components/spyglass/spyglass.js --fix` within this folder.

mgautam98:
That's actually great. We really needed it. I will ping you after making changes. Thanks

mgautam98:
@birm config file is giving this
```
ESLint couldn't find the config "google" to extend from. Please check that the name of the config is correct.

The config "google" was referenced from the config file in "/home/gautam/Projects/caMicroscope/.eslintrc.js"
```

birm:
In order to have it work on CI at least, we needed to install both eslint and eslint-config-google as non-global and use that binary. That is,
```
npm install eslint
npm install eslint-config-google
./node_modules/.bin/eslint core/*.js 
```

mgautam98:
@birm fixed the style.

PR #277, opened 2020-03-30
Nav-bar hover effect (with active page indicator) and icons addition

akhil-rana:
1. Active Section can now be clearly identified.
2. Hover effect on nav bar items.
3. Icons are added to the nav bar items.

ezgif-5-43389f8a62ae

### DIFF
--- a/apps/table.html
+++ b/apps/table.html
@@ -108,7 +108,7 @@
 				<div class="modal-content">
 					<div class="modal-header">
 						<h5 class="modal-title" id="title-of-dialog">Upload New Slide</h5>
-						<button type="button" class="close" onclick="initialize(); location.reload();" data-dismiss="modal" aria-label="Close">
+						<button type="button" class="close" onclick="initialize();" data-dismiss="modal" aria-label="Close">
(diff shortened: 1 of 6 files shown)


apps/table.html
```diff
@@ -96,16 +99,16 @@ <h1 class="h1">caMicroscope</h1>
 
 	<div class='p-2 bg-light' style="flex-grow: 1;">
 		<button type="button" class="btn btn-success float-right mb-2"
-			onclick="(()=>{location.reload();return false;})()">Reload</button>
+			onclick="(()=>{location.reload();return false;})()"> <i class="fas fa-sync-alt"></i> Reload</button>
```
mgautam98: It would be better if you change location.reload() to initialize().

akhil-rana: Yeah you are right. I did not see that. 
I did not add location.reload(), it was there already.
I think it should be changed though.

akhil-rana: I'll change that in the next PR if no one else does it.

akhil-rana:
Thank you

Issue #274, opened 2020-03-29
Finish Cleaning up style


birm:
To get the CI build to work, we've added some ignore rules to eslint. Let's work to remove as many as we feel are reasonable.

- [x]     "camelcase" : 0,
- [x]     "no-var" : 0,
- [x]     "prefer-promise-reject-errors" :0,
- [x]     "no-extra-bind" : 0,

cjchirag7:
@birm , may I work on this ? I feel that we can remove most of these.

cjchirag7:
@birm , we are still ignoring the 'no-var' rule , shall we remove this also ?

birm:
I think it's fine for now.

PR #281, opened 2020-03-30
Removed 'ignore no-extra-bind' in ESLint configuration

cjchirag7:
Fixes a sub-issue of #274
I have made sure that the application works, as earlier.
@birm , please review.


birm:
Thanks for looking into these style updates!

PR #283, opened 2020-03-30
adding login page

darkCavalier11:
the signup page has a link for login but nothing happened after the click. here I added a login page which opens when **already a user? login** is pressed

### DIFF
--- a/apps/signup/signup.css
+++ b/apps/signup/signup.css
@@ -10,22 +10,22 @@ body {
 	border-color: #e1e1e1;
 	border-radius: 0;
 }
-.signup-form {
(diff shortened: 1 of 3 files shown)


birm:
Right now, we're using an identity provider to handle authentication/login. I've re-added some somehow missing [docs to distro](https://github.com/camicroscope/Distro/tree/develop#securing-camicroscope) to try to describe this better. If you want to do this, you'll need to add another auth service to handle in-db credentials.

darkCavalier11:
can you pls describe how the login/authentication work. I want to contribute for further enhancement.

PR #284, opened 2020-03-30
Redirect bug fix

mgautam98:
@birm I found an interesting bug here
https://github.com/camicroscope/caMicroscope/blob/75561aff1c5c259b502774da8566e9e3d2d92318/apps/viewer/init.js#L1148
The setInterval runs atleast once before checking the conditions inside. It was causing exacly on second of delay after clicking on home button. 
After the fix it will redirect instantly.


mgautam98:
Thanks

PR #287, opened 2020-03-30
Client side file extension validaion

mgautam98:
File extension will be validated at client side also. Users will get feedback message as following.
Screenshot from 2020-03-31 03-35-36

Some in the slack group are saying that SlideLoader file extension validation is failing. So, just to be safe and also to improve user experience.

### DIFF
--- a/apps/table.html
+++ b/apps/table.html
@@ -25,7 +25,8 @@
 	<script src="./loader/loader.js"></script>
 	<script src="./loader/chunked_upload.js"></script>
 	<script>
-		var existing_slide_names = [];
+		var existingSlideNames = [];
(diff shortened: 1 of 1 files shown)


apps/table.html
```diff
@@ -36,11 +37,18 @@
...
+				if(existingSlideNames.includes(slide.value)) {
 					changeStatus("UPLOAD", "Slide with given name already exists");
 					return false;
 				}
 			}
+			//Checking for extinsion
+			let filename = document.getElementById("filename0").value
+			var fileExtension = filename.toLowerCase().split('.').reverse()[0];
+			if(!allowedExtensions.includes(fileExtension)) {
+				changeStatus("UPLOAD", fileExtension + " files are not allowed");
```
birm:
```suggestion
				changeStatus("UPLOAD", fileExtension + " files are not compatible");
```
A bit more descriptive?

mgautam98: Yes. Sure!

PR #288, opened 2020-03-30
Additional fixes to #284

mgautam98:
Hey @birm, I forgot to add theses to #284, Whithout these the page will not redirect to home after 5s. I am really sorry for that.

Thanks


birm:
Ah, I only did a minor smoke test, so I missed that too. Thanks for the fix to the fix :stuck_out_tongue:

PR #289, opened 2020-03-31
Check name duplicate

amritansh22:
In the segment app, when the name of the uploaded models are same, it creates confusion later while choosing the model and error while removing the model. This PR fixes these issues. 

So, now when we enter repeated name, we see an error as:</br></br>
image


amritansh22:
This PR had some non resolvable merge conflicts. It is referenced here #318

PR #303, opened 2020-04-03
Delete slide button in the Table

akhil-rana:
ezgif-4-539473bfbf2f


apps/table.html
```diff
@@ -446,6 +468,25 @@ <h3 class="text-center h3 mb-0">Available Slides</h3>
...
+		  });
+	});
+	function deleteSld(e) {
+		const oid = e.dataset.id;
+		const store = new Store('../data/');
+		if (oid) {
+			$('#deleteModal').modal('toggle');
+			$("#confirmDelete").unbind( "click" );
+			$("#confirmDelete").click(function(){
+				store.deleteSlide(oid);
```
mgautam98, line 483: Akhil I have implemented the same but here it returns a promise so, you can check if data.result.ok is 1 then initialize the table.

cjchirag7:
I think it would be better, if in the Modal we tell the user about the name and id of the slide, which he is deleting, so that he can make his confirmation to delete, without any doubt.

apps/table.html
```diff
@@ -446,6 +468,25 @@ <h3 class="text-center h3 mb-0">Available Slides</h3>
 		}
 	}
 
+	$(document).ready(function(){
+		$('#deleteModal').on('hidden.bs.modal', function (e) {
+			initialize();
+		  });
+	});
+	function deleteSld(e) {
+		const oid = e.dataset.id;
```
cjchirag7, line 477:
```suggestion
		const oid = e.dataset.id;
		const oname = e.dataset.name;		
```

apps/table.html
```diff
@@ -446,6 +468,25 @@ <h3 class="text-center h3 mb-0">Available Slides</h3>
...
+		const store = new Store('../data/');
+		if (oid) {
+			$('#deleteModal').modal('toggle');
```
cjchirag7, line 480:
```suggestion
			$('#confirmDeleteContent').html(`Are you sure you want to delete the slide ${oname} with id ${oid} ?`);
			$('#deleteModal').modal('toggle');
```

akhil-rana:
Thanks @cjchirag7

nanli-emory:
@cjchirag7 @mgautam98, I appreciate that you guys work hard for deleting slides.  The problem is that you must determine the current user who has the permission to delete the slide or not. It involves authority management. In the current situation, there are two types of users, one is `admin` who has the right to delete the slide. Another is `normal` who can't do delete operation. You guys need to add the authority check for the deleting functionality. What do you think @birm?

akhil-rana:
I think the route is configured to allow for Admin and Editor only.
Since we are using this particular route so.

https://github.com/camicroscope/Caracal/blob/master/caracal.js#L62

@birm  What are your thoughts?

birm:
Yes, in instances where we are checking attributes, slide deletion should only be performable by Admins and Editors (same people who can add a slide), and, as applicable, have a userFilter matching the slide's filter field (same people who can see the slide). I'm not sure it's foolproof yet, but that _should_ be taken care of by the backend.

cjchirag7:
We can do one thing that display the delete icon only, if the user is Editor or admin, so to avoid unnecessary requests to the backend also.

nanli-emory:
@cjchirag7 Thanks. This is what I'm talking about. From the UX perspective, we shouldn't show up a `delete` button but users actually can't use it.

akhil-rana:
@nanli-emory 
Would it work if the user gets an error pop up saying "Delete Permission not granted" on the button click?

akhil-rana:
Or we could send a dummy delete request on page initialisation just to check if the user has permission or not. 
and then disable the display button accordingly on page load itself.

birm:
You may notice a function called [parseJwt](https://github.com/camicroscope/caMicroscope/blob/24e00e5920b4c112280d3d1a8adeb8eba962bb68/common/util.js#L561)  which can get token data, which should include userType and userFilter, matching what's in the user mongo collection.

nanli-emory:
@akhil-rana We should get the `user type` from the `token`. show the `delete` button if `user type` is `admin`. Otherwise,  don't add the `delete` button on the page.
@birm, did you change the way to read `attr` from `token`?

akhil-rana:
@birm  Thanks for the info.
but in develop, we have NULL userType.
Who also has the delete permission(in case of DISABLE_SEC). If I disable the delete button then it'll be gone from develop branch? 
Is there any way to check the DISABLE_SEC value?

nanli-emory:
I will add the function called `getUserType` into `caMicroscope/common/util.js`.
This function will return `NULL` if the user is `normal` or no `token` (when turn off security). It will return `Admin` if the user has `token` and is `admin`.

akhil-rana:
Okay. Great !!
Thank you.

nanli-emory:
@akhil-rana  I added `getUserType()` that will return `string` value. 
admin -> `Admin` or normal/no security -> `Null`.

akhil-rana:
Thank you for your effort. 
I will commit the changes soon.

akhil-rana:
@nanli-emory  The function will always return Null when we work with develop.yml since we have no security in that case. 
So should the delete button be hidden on no security too? 
In that case anyone can delete the slide by going to the correct route anyway.

birm:
We may need to either add a route which displays some safe-to-show env variables the server has.

cjchirag7:
@nanli-emory , the code in util.js is not compiling, after the latest commit.
The word 'return' is misspelt. I have corrected it and fixed some lint issues in PR  #310

akhil-rana:
@birm Yes, I was thinking the same thing. 
What do you suggest us to do with this right now then?
Should I add a PR on Caracal for the same if you want?

birm:
Sure! I've added an issue matching what I think may be a scalable strategy.

akhil-rana:
@birm  @nanli-emory 
I have used the `getUserType` function to pass the userType as query parameter into the new `what can I do` route in Caracal. The delete button is will only be shown to the user to who has the permission to delete.

akhil-rana:
Thank you all for the help and effort.

PR #304, opened 2020-04-03
Feat Add Delete for Slides

mgautam98:
Added delete functionality for the slides


mgautam98:
Oops its already been added in #303 So closing it.

PR #313, opened 2020-04-04
Fix upload dialog and post only when check is successful [Fixes #286 ]

### LINKED ISSUE #286: UI Loader - only post if check successful
We have a number of slides added to camicroscope.org which are just broken, and I think this is the culprit.

cjchirag7:
Fixes #286 
The following features have been added :-
- Post button is shown only, if check is successful.
- When check button is clicked, first the form is validated. If valid, then only proceed to 'preview'.
- Close the upload-dialog automatically, when 'post' is successful
- Instead of loading the whole page again after posting, just call initialize() again 
- Show a popup message saying 'Slide uploaded successfully' for 2 seconds, when 'post' is successfully.

**Screencast**
slide-upload-new


mgautam98:
@cjchirag7 won't it make the "Finish upload" button effectively useless?

cjchirag7:
@mgautam98 , you are right. I forgot to take 'finish upload' into account while 'checking' for the status.
I working to show only 'Finish upload' to user, first, if it is successful, show 'Check', and if it is also successful, then only show the 'Post' button.
Will this be fine ?

birm:
I'm sorry for my confusion, but what about this invalidates the finish upload functionality?

cjchirag7:
I don't understand how it invalidates the finish upload functionality. I have not made any change 'Finish Upload' functionality in this PR :confused:  . Can you please elaborate ?

cjchirag7:
@birm , I have made the following changes, in previous 4 commits :
- 'Check' button would only be shown if 'FinishUpload' is successful.
- 'Post' button would only be shown if both 'FinishUpload' and 'Check' are successful.
- If after checking, user changes the file or the FileName, he will see only 'FinishUpload' again, and he needs to follow the process of 'Check' again, inorder to post the slide with changes.
- Earlier, the modal for 'Slide Upload' closed, whenever user clicked outside the modal, but now, the modal closes only when 'Close' button is clicked or 'x' button on top is clicked.
- When the modal is closed, the form is reset.
- Instructions for Uploading a slide are updated.
- Allow only single file to be uploaded at a time (Because, the functions validateForm() , sanitize() etc. were validating only the first file uploaded. May be, we can extend this facility to multiple files in another PR, if it is required. )

slide-upload

apps/table.html
```diff
@@ -122,9 +144,9 @@ <h1 class="h1">caMicroscope</h1>
 		<button type="button" class="btn btn-success float-right mb-2"
 			onclick="(()=>{location.reload();return false;})()"> <i class="fas fa-sync-alt"></i> Reload</button>
 		<button type="button" class="btn btn-primary float-right mb-2 mr-2" data-toggle="modal"
-			data-target="#upload-dialog"> <i class="fas fa-upload"></i> Upload</button>
+			data-target="#upload-dialog" onclick="hidePostButton(); hideCheckButton(); resetUploadForm();"" <i class="fas fa-upload"></i> Upload</button>
```
akhil-rana, line 147: @cjchirag7 
You missed a closing tag here (>), and also a extra quote is present( " ) due to which the upload icon is no longer visible.

birm: on it.

cjchirag7:
Oops! Thanks for pointing it out and correcting it.

PR #317, opened 2020-04-04
fix segment redirect and refactored segment.html

mgautam98:
For invalid slideId or no slideId It slould be redirected back to the table pages but it was again redirected to same page.

ezgif com-video-to-gif (1)

Fixed the problem also put the css to a different file to make it easily understandable and able to debug.


mgautam98:
there is same redirect problem with model also I will do the changes

birm:
If you'd like, you can put the model change in this PR too :)

mgautam98:
@birm made the changes for model also

PR #318, opened 2020-04-04
Check name duplicate

amritansh22:
In the segment app, when the name of the uploaded models are same, it creates confusion later while choosing the model and error while removing the model. This PR fixes these issues. 

So, now when we enter repeated name, we see an error as:</br></br>
image


apps/model/model.js
```diff
@@ -154,6 +154,13 @@ async function initUIcomponents() {
     headerText: 'Help',
     hasFooter: false,
   });
+  // Create Modal to take input from user of new class list
+  $UI.chngClassLst = new ModalBox({
```
mgautam98, line 158: chngClassLst UI component has already been defined just above that. It's a mistake I think.

mgautam98:
@amritansh22 it is not working if I refresh the page remember that the models are persistent in the memory therefore, this feat. also needs to be persistent.

Also, it is not working if I don't refresh, it still uploads the model becaue you need to return after catching the error. Here, it displays it and then continues the execution.

apps/model/model.js
```diff
@@ -671,6 +686,16 @@ function uploadModel() {
           return;
         }
 
+        try {
+          if (checkNameDuplicate(_name.value)) {
+            throw new Error('Model name repeated');
+          }
+        } catch (e) {
+          status.innerHTML = 'Model with the same name already exists. Please choose a new name';
+          status.classList.remove('blink');
```
mgautam98, line 695: Needs to return form here else it will still save the model.

apps/model/model.js
amritansh22: Thanks for pointing it out. It was by mistake.

apps/model/model.js
amritansh22: Thanks for pointing out. I have fixed it.

amritansh22:
@mgautam98 Thanks a lot for pointing out the mistakes. I have fixed them.

mgautam98:
@birm are you sure it's working. The namearr is not being populated with all the model names. Though, model name is saved as soon as you refresh or go to other pages the namearr is again initialised to empty array.

birm:
Hm, I may have missed something when testing... If there's a a fix to this fix, we'll certainly welcome a PR.

PR #322, opened 2020-04-05
Store gradient/color information in DB [ Fixes #307 ]

### LINKED ISSUE #307: Heatmap Updates don't seem to store color/gradient info.
**Describe the bug**
Editing heatmap colors/bins does not reflect, instead only two bins of green are seen.

**To Reproduce**
Edit a heatmap and its colors

**Expected behavior**
Data in DB reflects color/bin changes

update data query body like: ```{"provenance.analysis.fields":[{"name":"necrosis","range":[0,1],"value":[0,1]},{"name":"tumor","range":[0,10],"value":[0,10]}],"provenance.analysis.setting":{"mode":"gradient","field":"necrosis"}}```

cjchirag7:
Fixes #307 
To store the colors , provenance.analysis.setting will have a new field 'colors'
- When mode is 'binal', length of colors is 1
- When mode is 'gradient', length of colors is >=2

The query which was earlier something like -
```
{"provenance.analysis.fields":[{"name":"necrosis","range":[0,1],"value":[0,1]},{"name":"tumor","range":[0,10],"value":[0,10]}],"provenance.analysis.setting":{"mode":"gradient","field":"necrosis"}}
```
is modified to -
```
{"provenance.analysis.fields":[{"name":"necrosis","range":[0,1],"value":[0,1]},{"name":"tumor","range":[0,10],"value":[0,10]}],"provenance.analysis.setting":{"mode":"gradient","field":"necrosis",colors":["#F23112","#553172","#333117" ] }}
```

In case, when mode is 'binal', the value of `color[0]` is the selected color .

### DIFF
--- a/apps/heatmap/init.js
+++ b/apps/heatmap/init.js
@@ -128,18 +128,11 @@ function initCore() {
 
         if ($D.heatMapData.provenance.analysis.setting) {
           opt.mode = $D.heatMapData.provenance.analysis.setting.mode;
-          if (
-            opt.mode === 'binal' &&
-            $D.heatMapData.provenance.analysis.setting.color
-          ) {
-            opt.color = $D.heatMapData.provenance.analysis.setting.color;
-          } else if (
-            opt.mode === 'gradient' &&
-            $D.heatMapData.provenance.analysis.setting.colors &&
-            $D.heatMapData.provenance.analysis.setting.colors.length>0
-          ) {
+          if (opt.mode === 'binal') {
+            opt.color = $D.heatMapData.provenance.analysis.setting.colors[0];
+          } else if (opt.mode === 'gradient') {
             opt.colors = $D.heatMapData.provenance.analysis.setting.colors;
-            opt.steps = $D.heatMapData.provenance.analysis.setting.colors.length+1;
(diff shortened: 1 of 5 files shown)


cjchirag7:
**A Screencast of the working**
store-color-update-heatmap

apps/heatmap/init.js
```diff
@@ -123,11 +123,24 @@ function initCore() {
...
+          ) {
+            opt.color = $D.heatMapData.provenance.analysis.setting.color;
+          } else if (
+            opt.mode === 'gradient' &&
+            $D.heatMapData.provenance.analysis.setting.colors &&
+            $D.heatMapData.provenance.analysis.setting.colors.length>0
+          ) {
+            opt.colors = $D.heatMapData.provenance.analysis.setting.colors;
+            opt.steps = $D.heatMapData.provenance.analysis.setting.colors.length+1;
+          }
```
birm, line 143: Would it be reasonable to just use "colors" as a db field for both cases, and just an array of len 1 in binial case?

cjchirag7: Yeah, it would be better. I would update the code.

cjchirag7: I have updated the code. Please review once.

Issue #116, opened 2018-04-30
Use templates properly


sharmaashish:
Our sliders should be templated. Here is an example of code that should have been templated. During our refactoring, we should be able to push the template into Mongo, and then populate the form from a separate file that relies on that template.
https://github.com/camicroscope/caMicroscope/blob/e7c690e140efdd1730352827ea3139508bae5c4e/js/annotationtools/osdAnnotationTools.js#L2414-L2622

birm:
This is referring to templatizing dynamic services interactions, for future reference.

birm:
See https://github.com/camicroscope/caMicroscope/blob/refactor/js/dataloaders.js; we're not hardcoding any templates with the refactor.

PR #323, opened 2020-04-05
Initial Implementation of Information Dashboard

Hemansh31:
Fixes https://github.com/camicroscope/Distro/issues/95
This is an initial implementation of the Information Dashboard.
image
**Few Ideas**
Clicking on each slide in the table could open a popup with all the summary related to the slide.
@birm Please suggest what else to include.

### DIFF
--- a/core/Store.js
+++ b/core/Store.js
@@ -171,11 +171,11 @@ class Store {
       mode: 'cors',
     }).then(this.errorHandler).then((x)=>this.filterBroken(x, 'mark'));
   }
-  fetchMark(slide_id) {
+  fetchMark(slideId) {
     const suffix = 'Mark/find';
     const url = this.base + suffix;
     const query = {};
-    query['provenance.image.slide'] = slide_id;
+    query['provenance.image.slide'] = slideId;
     return fetch(url + '?' + objToParamStr(query), {
       credentials: 'include',
       mode: 'cors',
@@ -264,11 +264,11 @@ class Store {
       mode: 'cors',
     }).then(this.errorHandler).then((x)=>this.filterBroken(x, 'heatmap'));
   }
-  fetchHeatMap(slide_id) {
+  fetchHeatMap(slideId) {
     const suffix = 'Heatmap/find';
     const url = this.base + suffix;
     const query = {};
-    query['provenance.image.slide'] = slide_id;
+    query['provenance.image.slide'] = slideId;
     return fetch(url + '?' + objToParamStr(query), {
       credentials: 'include',
       mode: 'cors',
--- a/core/Store.js
+++ b/core/Store.js
@@ -171,6 +171,16 @@ class Store {
       mode: 'cors',
     }).then(this.errorHandler).then((x)=>this.filterBroken(x, 'mark'));
   }
+  fetchMark(slide_id) {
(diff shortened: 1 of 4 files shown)


birm:
Oh, cool; just looks like you have a minor style fix, and you're good :)

Hemansh31:
@birm what style fix? 
and should I also implement the popup feature I mentioned now.

core/Store.js
```diff
@@ -171,6 +171,16 @@ class Store {
       mode: 'cors',
     }).then(this.errorHandler).then((x)=>this.filterBroken(x, 'mark'));
   }
+  fetchMark(slide_id) {
```
birm, line 174:
```suggestion
  fetchMark(slideId) {
```

core/Store.js
```diff
@@ -171,6 +171,16 @@ class Store {
...
+    const suffix = 'Mark/find';
+    const url = this.base + suffix;
+    const query = {};
+    query['provenance.image.slide'] = slide_id;
```
birm, line 178:
```suggestion
    query['provenance.image.slide'] = slideId;
```

core/Store.js
```diff
@@ -254,6 +264,16 @@ class Store {
       mode: 'cors',
     }).then(this.errorHandler).then((x)=>this.filterBroken(x, 'heatmap'));
   }
+  fetchHeatMap(slide_id) {
```
birm, line 267:
```suggestion
  fetchHeatMap(slideId) {
```

core/Store.js
```diff
@@ -254,6 +264,16 @@ class Store {
...
+    const suffix = 'Heatmap/find';
+    const url = this.base + suffix;
+    const query = {};
+    query['provenance.image.slide'] = slide_id;
```
birm, line 271:
```suggestion
    query['provenance.image.slide'] = slideId;
```

Hemansh31:
I mean Clicking on each slide in the table could open a popup with all the summary related to the slide.

birm:
That sounds reasonable to me!

birm:
It's up to you if you want us to merge this for now, and do that later, or include that as part of the same PR

Hemansh31:
@birm I prefer to include the feature as a new PR, you could merge this PR for now

Vedant1202:
Hi @Hemansh31, 
The solution you've provided is a good quick fix to the feature request and even occured to me at the first instance, however, it would be very heavy on the server when scaled. This is because you are making a request for the heatmap and annotations seperately for each slide.
That means if you have about 10 slides in a db, you will be making 10 request for heatmaps and 10 requests for annotations, i.e., 20 requests in total. Additionally these 20 requests would be hitting the database everytime. 
So, every reload on the page containing 10 slides makes 40 requests in total which is frankly a very bad ratio. This would cause the server to hang up or even crash if scaled.
A good idea is to gather related data at the database layer itself, so that the data needed gets fetched in a single request as I've done here [#332](https://github.com/camicroscope/caMicroscope/pull/332). This also includes the details modal with you and @birm were discussing about. 
A catch in this fix is that the solution requires mongo 4.0 and up to be implemented. I've discussed the same with @birm and hence I've created a corresponding pull request at [#116](https://github.com/camicroscope/Distro/pull/116).

PR #325, opened 2020-04-06
Upload form redesign and  upload button visibility

akhil-rana:
1. The Slide Upload form is slightly redesigned to match the application.
2. Now, the upload button will only be visible to the user who has the permission to upload a slide.(Using `what can i do` route)

Old:
Screenshot from 2020-04-06 07-07-38

New:
Screenshot from 2020-04-06 07-07-12


akhil-rana:
I also noticed that the `multiple` keyword was missing in the previous design itself.
I realised that,
It was removed in #313

birm:
Ok, then we can merge this and let simplifying the code be a separate issue.

Hemansh31:
After this merge with the build as instructed in the Readme the Upload button is not visible by default?
@akhil-rana how to correct this.
image

PR #327, opened 2020-04-06
fix duplicate model name

mgautam98:
Makes #318 working. 
Though, #318 says it fixes segmentation app but code works on predict app. Both the apps have same problem so, fixed both of them.


apps/segment/segment.js
```diff
@@ -584,6 +588,18 @@ function uploadModel() {
         var modelInput = tf.io.browserFiles([topology.files[0], ...weights.files]);
       }
 
+      //Check if model with same name is previously defined
```
birm, line 591: should fix the style build
```suggestion
      // Check if model with same name is previously defined
```

PR #331, opened 2020-04-06
Clean loader code and increase upload modal width

### LINKED ISSUE #326: Clean up Loader Code from Multiple Concurrent Uploads
We can simplify the loader code now that we're only supporting one slide at a time.

cjchirag7:
Closes #326 
This includes the following changes :
- Refactored loader code to support only a single file upload, at a time
- Cursor has been made :point_up_2: when users hovers over the 'choose file' (browse) button in the modal
- The resetUploadForm() has been modified, in accordance with #325 
- The upload modal width is increased
After increased width, the modal look like this, on an average PC -
wide-modal

### DIFF
--- a/apps/loader/chunked_upload.js
+++ b/apps/loader/chunked_upload.js
@@ -69,25 +69,22 @@ async function handleUpload(selectedFiles) {
     idtr.deleteCell(1);
   }
 
-  var currID = 0;
   // Add columns
-  for (var i=0; i<selectedFiles.length; i++, currID++) {
-    idtr.insertCell(-1).innerHTML = '<b>'+Number(currID+1)+'<b>';
-    fnametr.insertCell(-1).innerHTML = `<input type=text class="form-control" name=filename id='filename${currID}'
-    onchange=hideCheckButton();hidePostButton(); value='${selectedFiles[i]['name']}'>`;
-    tokentr.insertCell(-1).innerHTML = `<input type=text class="form-control" onchange=hideCheckButton();hidePostButton();
-    disabled name=token id='token${currID}'>`;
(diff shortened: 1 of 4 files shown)


apps/loader/chunked_upload.js
```diff
@@ -69,25 +69,22 @@ async function handleUpload(selectedFiles) {
...
   // Add columns
-  for (var i=0; i<selectedFiles.length; i++, currID++) {
-    idtr.insertCell(-1).innerHTML = '<b>'+Number(currID+1)+'<b>';
-    fnametr.insertCell(-1).innerHTML = `<input type=text class="form-control" name=filename id='filename${currID}'
-    onchange=hideCheckButton();hidePostButton(); value='${selectedFiles[i]['name']}'>`;
-    tokentr.insertCell(-1).innerHTML = `<input type=text class="form-control" onchange=hideCheckButton();hidePostButton();
-    disabled name=token id='token${currID}'>`;
-    slidetr.insertCell(-1).innerHTML = `<input type=text class="form-control" name=slidename id='slidename${currID}'>`;
-    filtertr.insertCell(-1).innerHTML = `<input type=text class="form-control" name=filter id='filter${currID}'>`;
+  idtr.insertCell(-1).innerHTML = '<b>'+Number(1)+'<b>';
```
birm, line 73: I think we should get rid of the ID field entirely. At worst, it may confuse people that the new slide's uuid is just 1.

cjchirag7: Sure, I will do this.

cjchirag7:
@birm , I have made the required changes.

Issue #337, opened 2020-04-07
Slide table not refreshed when uploading first slide


Vedant1202:
**Describe the bug**
When we upload the first slide, the table containing slides is not being seen. 

**To Reproduce**
1. Clear the entire database.
2. Go to upload slides page.
3. Upload a slide.

**Expected behavior**
The table should be seen containing the uploaded slides. This is not being seen because the table does not exist in the DOM and we've not added a condition to add it.

**Screenshots**
1. Clear database and go to upload slide
![Capture1](https://user-images.githubusercontent.com/35070972/78649955-d4b4a180-78db-11ea-9e78-587327b4ac37.PNG)
2. Upload a slide
![Capture2](https://user-images.githubusercontent.com/35070972/78649993-eac26200-78db-11ea-965b-89690b3c58ae.PNG)
3. No table visible.
![Capture3](https://user-images.githubusercontent.com/35070972/78650113-15141f80-78dc-11ea-944f-fdd116becc65.PNG)

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser Firefox
 - Version 72.0.2

Vedant1202:
This is fixed in PR [#338 ](https://github.com/camicroscope/caMicroscope/pull/338).

cjchirag7:
Actually, if you look in the browser console, this error is shown -
```
table.html:425 Uncaught (in promise) RangeError: Invalid array length
    at table.html:425
```

birm:
Ah, this was my bad. Thanks for the fix(es)!

PR #339, opened 2020-04-07
Fix for refresh table on first slide upload

Vedant1202:
A fix to [#337](https://github.com/camicroscope/caMicroscope/issues/337). 

This fix includes:
1.  Refresh page on first slide upload.
2. Check to make sure page does'nt refresh on concurrent slide uploads.


cjchirag7:
@Vedant1202 , have you pulled the latest code from the **develop** branch of caMicroscope repo ?
Some of your fixes  like fixing the typo for 'return', fixing lint issues, has already been made earlier in #310

cjchirag7:
@Vedant1202 , And why is it needed to refresh table on the first slide upload separately ? In #313 , I have updated the code to call initialize(), when 'post' of slide is successful, instead of reloading the whole page again.
One more thing, if you reload the whole page again, instead of calling initialize(), the popup showing 'Slide uploaded successfully' (added in #313 ) , would not work correctly.

Vedant1202:
@cjchirag7 Yes, I have the latest pull from develop branch and the typo issue was just a merge fix, so it popped up.

Vedant1202:
@cjchirag7 The initialize is not working during first slide upload. This is because it depends on a DOM element '#datatables' which does not exist in the first place (refer this [line](https://github.com/camicroscope/caMicroscope/blob/8b9bfc87704858d6b24fd5d59424569a7463960a/apps/table.html#L418)). 
A better fix to refreshing would be putting a condition to add the DOM element if it does not exist. 

Also I've kept in mind that the refresh is not interfering with the popup, as it is triggerred only after the popup gets hidden.

cjchirag7:
@Vedant1202 , I noticed this. I have fixed in such a way that it adds the DOM element $('datatables) , when it doesn't exist, in #340

Vedant1202:
Yeah, works

PR #340, opened 2020-04-07
Fix initialise() on first slide upload

cjchirag7:
I have made the fix such that, it adds the DOM element $('datatables) , when it doesn't exist, which was the major cause of this problem.
**Screencast of the working**
auto-refresh-slides


cjchirag7:
Closes #337

Issue #324, opened 2020-04-06
File already exists error even after the slide is deleted.


akhil-rana:
**Describe the bug**
`File with name 'sample.svs' already exists` error is shown even if the file was previously deleted from the slide table

**To Reproduce**
1. Upload a slide
2. Delete the same slide from slides table
3. Now try to re upload the same file without changing file name.

**Screenshots**
![ezgif-5-6d66c347e9ab](https://user-images.githubusercontent.com/36575628/78514869-4227db80-77d1-11ea-97f9-464956f3e27c.gif)

**Desktop :**
 - OS: Ubuntu 
 - Browser: Chrome

cjchirag7:
I think, when deleting a slide, the file is not deleted at the backend, only the data is deleted from the DB. Due to this, this might happen.

akhil-rana:
Yes, I think so too.
@birm Any thoughts?

birm:
Yes. There's a separation between slides existing on the fs and being registered in the db. Our slide process is pretty confusing.

akhil-rana:
Okay, any suggestions on how to delete from fs?

Vedant1202:
Hi @birm and @akhil-rana, I think [#341 ](https://github.com/camicroscope/caMicroscope/pull/341) and [#25](https://github.com/camicroscope/SlideLoader/pull/25) should fix this.

cjchirag7:
I think [#23](https://github.com/camicroscope/Caracal/pull/23) is a better fix, as it takes care that when the slide is deleted from the DB, then only the file is deleted locally, and keeps the route secure as earlier.
[#25](https://github.com/camicroscope/SlideLoader/pull/25) has some major problems with the route security, which I have commented on the PR. These should be looked carefully, before performing any merge.

akhil-rana:
@birm 
Should this be closed?

PR #341, opened 2020-04-07
Functionality to delete slide from file system

Vedant1202:
This fix contains: 

1. Call to delete slide from file system.
2. Appropriate alerts and callbacks for successful deletion of file.
3. Update to deleteSlide function by chaining it to 'delete from file system' function.

This fix solves [#324 ](https://github.com/camicroscope/caMicroscope/issues/324).


cjchirag7:
@Vedant1202 , it would be better that the file is deleted using the node backend only (Caracal.js) and only when the delete from MongoDB database is successful. I have made this feature, and just testing it.

Vedant1202:
@cjchirag7, I've already considered this issue and have chained the processes to each other by promises. So, unless one promise is handled successfully, the other does'nt get executed. 

I've looked at your PR and there are some serious issues in it please look into it [here](https://github.com/camicroscope/Caracal/pull/23#issuecomment-610522407).

Vedant1202:
Also, putting all file access code in SlideLoader is much better just to make sure all related code stays in one place.

cjchirag7:
@Vedant1202 , yeah I didn't notice you're chaining the promises correctly. But your PR [#25](https://github.com/camicroscope/SlideLoader/pull/25) has some major problems with the route security, which I have commented on the PR. These should be looked into, carefully.
I have fixed all those issues with  [#23](https://github.com/camicroscope/Caracal/pull/23) . You may review the code once :)

Vedant1202:
@cjchirag7 I did the updates in [PR-25](https://github.com/camicroscope/SlideLoader/pull/25). You can have a look.

birm:
I think this looks good; just will have to wait to co-test this with caMicroscope/SlideLoader#25

apps/table.html
```diff
@@ -661,21 +756,36 @@ <h3 class="text-center h3 mb-0">Available Slides</h3>
...
+
 		const store = new Store('../data/');
 		if (oid) {
-			$('#confirmDeleteContent').html(`Are you sure you want to delete the slide ${oname} with id ${oid} ?`);
+			$('#confirmDeleteContent').html(`Are you sure you want to ${reqId ? 'cancel the ': ''} ${getUserType() === 'Admin' ? '' : 'request to ' } delete the slide ${oname} with id ${oid} ?`);
 			$('#deleteModal').modal('toggle');
 			$("#confirmDelete").unbind( "click" );
 			$("#confirmDelete").click(function(){
-				store.deleteSlide(oid);
+				if (userType === "Admin" && !cancel) {
```
akhil-rana, line 775: @Vedant1202 @birm 
I think that delete permission should also be allowed in case of DISABLE_SEC (no security)
And since usertype is "null" in that case, I think instead of using `getUserType` here, `getUserPermissions` maybe be more preferable since it uses `wcido`

birm: I think that's right. Want to submit a PR?

birm: (you could also extend #367 to include that, since it's so similar)

akhil-rana: > (you could also extend #367 to include that, since it's so similar)

Okay, I'll work on it.

Vedant1202: Yeah, that works

PR #345, opened 2020-04-07
Download button in the slides table

akhil-rana:
A download button is added in the slides table which the user to download the slide image.

This is used with [SlideLoader #26](https://github.com/camicroscope/SlideLoader/pull/26)

Screenshot from 2020-04-08 03-24-40

### DIFF
--- a/apps/loader/loader.js
+++ b/apps/loader/loader.js
@@ -110,8 +110,7 @@ function handleUpload(file, filename) {
   );
 }
 
-function handleDownload(e) {
-  var id=e.dataset.id;
+function handleDownload(id) {
   var fileName='';
   store.getSlide(id)
       .then((response) => {
--- a/apps/loader/loader.js
+++ b/apps/loader/loader.js
@@ -114,7 +114,9 @@ function handleDownload(id) {
   var fileName='';
   store.getSlide(id)
       .then((response) => {
-        return response[0]['location'];
+        if (response[0]['location']) {
+          return response[0]['location'];
+        }
       }).then((location) => {
         fileName= location.substring(8, location.length);
         return fileName;
@@ -122,7 +124,13 @@ function handleDownload(id) {
         fetch(downloadURL + fileName, {
           credentials: 'same-origin',
           method: 'GET',
-        }).then((response) => response.blob())
+        }).then((response) => {
+          if (response.status == 404) {
+            throw response;
+          } else {
+            return response.blob();
+          }
+        })
             .then((blob) => {
               var url = window.URL.createObjectURL(blob);
               var a = document.createElement('a');
@@ -137,6 +145,8 @@ function handleDownload(id) {
               alert('Error! Can\'t download file.');
             },
             );
+      }).catch((error) => {
+        console.log(error);
       });
 }
 
--- a/apps/loader/loader.js
+++ b/apps/loader/loader.js
@@ -1,6 +1,7 @@
 var uploadUrl = '../loader/upload/start/';
 var checkUrl = '../loader/data/one/';
 var thumbUrl = '../loader/data/thumbnail/';
+var downloadURL = '../loader/getSlide/';
 
 var store = new Store('../data/');
 
@@ -108,6 +109,38 @@ function handleUpload(file, filename) {
       (error) => changeStatus('UPLOAD', error), // Handle the error response object
   );
 }
+
+function handleDownload(e) {
+  var id=e.dataset.id;
+  var fileName='';
+  store.getSlide(id)
+      .then((response) => {
+        return response[0]['location'];
+      }).then((location) => {
+        fileName= location.substring(8, location.length);
+        return fileName;
+      }).then((fileName) =>{
+        fetch(downloadURL + fileName, {
+          credentials: 'same-origin',
+          method: 'GET',
+        }).then((response) => response.blob())
+            .then((blob) => {
+              var url = window.URL.createObjectURL(blob);
+              var a = document.createElement('a');
+              a.href = url;
+              a.download = fileName;
+              document.body.appendChild(a);
+              a.click();
+              a.remove(); // afterwards we remove the element again
+              window.URL.revokeObjectURL(blob);
+            }).catch((error) =>{
+              console.log(error);
+              alert('Error! Can\'t download file.');
+            },
+            );
+      });
+}
+
 
 function handleCheck(filename, reset, id) {
   fetch(checkUrl + filename, {credentials: 'same-origin'}).then(
--- a/apps/loader/loader.js
+++ b/apps/loader/loader.js
@@ -120,7 +120,8 @@ function handleDownload(id) {
           throw new Error('Slide not found');
         }
       }).then((location) => {
-        fileName= location.substring(8, location.length);
+        fileName= location.substring(location.lastIndexOf('/')+1, location.length);
+        console.log(fileName);
         return fileName;
       }).then((fileName) =>{
         fetch(downloadURL + fileName, {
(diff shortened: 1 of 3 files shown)


apps/table.html
```diff
@@ -372,7 +372,9 @@ <h5 class="modal-title" id="deleteModalLabel">Delete confirmation</h5>
 							else if (!d[key]) rs.push('')
 							else rs.push(d[key])
 						});
-						const btn = `<div id='open-delete'> <button class="btn btn-success" data-id='${rs[0]}' onclick='openView(this)'>Open</button> <button type='button' class='btn btn-danger DelButton' id='deleteBtn' data-id='${rs[0]}' data-name='${rs[1]}' onclick='deleteSld(this)' data-toggle='modal'> <i class='fas fa-trash-alt' ></i> </button></div>`
+						const btn = `<div id='open-delete'> <button class="btn btn-success" data-id='${rs[0]}' onclick='openView(this)'>Open</button> 
+							<button type='button' class='btn btn-danger DelButton' id='deleteBtn' data-id='${rs[0]}' data-name='${rs[1]}' onclick='deleteSld(this)' data-toggle='modal'> <i class='fas fa-trash-alt' ></i> </button>
+							<button type='button' class='btn btn-info DownloadButton' id='downloadBtn' data-id='${rs[0]}' onclick='handleDownload(this)' > <i class='fas fa-download' ></i> </button></div>`
```
cjchirag7:
```suggestion
							<button type='button' class='btn btn-info DownloadButton' id='downloadBtn' data-id='${rs[0]}' onclick='downloadSlide(this)' > <i class='fas fa-download' ></i> </button>
							<button type='button' class='btn btn-danger DelButton' id='deleteBtn' data-id='${rs[0]}' data-name='${rs[1]}' onclick='deleteSld(this)' data-toggle='modal' style='display:none'> <i class='fas fa-trash-alt' ></i> </button> </div>`
```
These 2 changes can be made 
- The delete icon should not be displayed initially, as we are calling checkUserPermissions() in line 379 of table.html , which will automatically change the 'display' to 'block', iff the user has apt. permissions.
- The order of the download icon and delete icon can be reversed, so that the delete icons get shifted to the rightmost

apps/loader/loader.js
```diff
@@ -109,6 +110,37 @@ function handleUpload(file, filename) {
...
+      .then((response) => {
+        return response[0]['location'];
+      }).then((location) => {
+        fileName= location.substring(8, location.length);
+        return fileName;
+      }).then((fileName) =>{
+        fetch(downloadURL + fileName, {
+          credentials: 'same-origin',
+          method: 'GET',
+        }).then((response) => response.blob())
```
birm, line 125: interesting consequence, the client will download the error message as a html file. I guess we have to explicitly check for an error status code?

cjchirag7: @birm , yeah it would be required. 
And also, I noticed then we a user tries do delete a slide which doesn't exist (try this by calling handleDelete() with an id of a slide already deleted, then this error is given in console : 
`Uncaught (in promise) TypeError: Cannot read property 'location' of undefined
    at loader.js:117`
So, I would also request making a similar change in getSlide() function, to check the error status code, as well as if response[0] has a property 'location' or not.

apps/loader/loader.js
```diff
@@ -114,15 +114,23 @@ function handleDownload(id) {
   var fileName='';
   store.getSlide(id)
       .then((response) => {
-        return response[0]['location'];
+        if (response[0]['location']) {
+          return response[0]['location'];
+        }
```
cjchirag7, line 119:
```suggestion
        } else {
           throw new Error('Slide not available');
        }
```

akhil-rana: I also added something like this before but,
This `else` might not even execute because the error will be thrown in the fetch itself. Correct me if I am wrong.

cjchirag7, line 119: @akhil-rana 
This is for the case when response[0] is undefined, accessing response[0]['location'] would give a reference error.
```suggestion
        if (response[0]) {
          return response[0]['location'];
        }
        else 
        {
         throw new Error('Slide not found');
        }
```

akhil-rana: I'm sorry but I don't understand that when will `response[0]` be undefined.
In case of `id` can't be found, error will be thrown in the fetch itself so.

akhil-rana: Okay, I got it. 
Thanks for the changes.

cjchirag7: Actually, the response is an empty array, when id of a non-existing slide is provided, in handleDownload()
*Screenshot*
Screenshot from 2020-04-08 12-56-27

akhil-rana: Yes, I somehow missed that. Thanks.

apps/table.html
```diff
@@ -372,7 +372,9 @@ <h5 class="modal-title" id="deleteModalLabel">Delete confirmation</h5>
 							else if (!d[key]) rs.push('')
 							else rs.push(d[key])
 						});
-						const btn = `<div id='open-delete'> <button class="btn btn-success" data-id='${rs[0]}' onclick='openView(this)'>Open</button> <button type='button' class='btn btn-danger DelButton' id='deleteBtn' data-id='${rs[0]}' data-name='${rs[1]}' onclick='deleteSld(this)' data-toggle='modal'> <i class='fas fa-trash-alt' ></i> </button></div>`
+						const btn = `<div id='open-delete'> <button class="btn btn-success" data-id='${rs[0]}' onclick='openView(this)'>Open</button> 
+							<button type='button' class='btn btn-info DownloadButton' id='downloadBtn' data-id='${rs[0]}' onclick='downloadSlide(this)' > <i class='fas fa-download' ></i> </button>
+							<button type='button' class='btn btn-danger DelButton' id='deleteBtn' data-id='${rs[0]}' data-name='${rs[1]}' onclick='deleteSld(this)' data-toggle='modal'> <i class='fas fa-trash-alt' ></i> </button></div>`
```
cjchirag7, line 377: @akhil-rana , can you do one more favour ? 
Please add this 'display:none' style to the delete button initially, so as we have already been calling checkUserPermissions() in line 379 of table.html , which will automatically change the 'display' to 'block', iff the user has apt. permissions. 
Currently, the delete button is being displayed to all the users, irrespective of whether they have permission or not. 
This change will help fix this bug, in this PR itself -
```suggestion
							<button type='button' class='btn btn-danger DelButton' style='display:none;' id='deleteBtn' data-id='${rs[0]}' data-name='${rs[1]}' onclick='deleteSld(this)' data-toggle='modal'> <i class='fas fa-trash-alt' ></i> </button></div>`
```

akhil-rana: It is already set to `none` in `table.css` here [table.css](https://github.com/camicroscope/caMicroscope/blob/develop/apps/table.css#L51)

cjchirag7: Oh! my bad.

akhil-rana:
@birm 
Requested changes have been made.
Please review.

apps/loader/loader.js
```diff
@@ -109,6 +110,49 @@ function handleUpload(file, filename) {
...
+  var fileName='';
+  store.getSlide(id)
+      .then((response) => {
+        if (response[0]) {
+          return response[0]['location'];
+        } else {
+          throw new Error('Slide not found');
+        }
+      }).then((location) => {
+        fileName= location.substring(8, location.length);
```
birm, line 123: wouldn't it make more sense to split on "/" instead?

akhil-rana: Yes, of course. 
I'm on it.

akhil-rana: Looks right now?

akhil-rana:
Thanks for the help and suggestions.
@birm  @cjchirag7

PR #346, opened 2020-04-08
Feature drawing pen thickness

mgautam98:
The draw sub tool will include a feature to choose thickness for the drawing pen, only for free mode of drawing. Some users can choose it to make lines more visible.
Annotation 2020-04-08 114509

The users can choose the thickness from x1 to x8, where x1 is the original thickness. Default value for thickness will be x1.
Annotation 2020-04-08 114534

@nanli-emory @birm please review. 
Later I'll add the animation to the slider so, thumb size increases as the thickness is increased.

ezgif com-video-to-gif (5)

### DIFF
--- a/components/simplecontextmenu/simplecontextmenu.css
+++ b/components/simplecontextmenu/simplecontextmenu.css
@@ -178,5 +178,49 @@ height:102px;
   text-align: center;
 }
 
+.slider {
+  -webkit-appearance: none;
+  background: transparent;
+  outline: none;
+  opacity: 0.7;
+  -webkit-transition: .2s;
+  transition: opacity .2s;
+}
 
+.slider:hover {
+  opacity: 1;
+}
 
+.slider:focus{
+  outline: none;
+}
+
+.slider::-webkit-slider-thumb {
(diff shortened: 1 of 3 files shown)


birm:
So, just to be clear, this only impacts the visibility of the line while drawing it, right? It doesn't seem to alter the annotation saved.
I think that's ok.

mgautam98:
No, it will not alter the annotations saved. 

I didn't need to change anything for area and circumference calculation as by default it was taking outer border points on line.

mgautam98:
@birm do you want me to change the annot saved too?

birm:
I'm kind of torn on this one. I think it makes sense not to save thickness, as either way to interact with lines on a zoomable image may not be intuitive.
I'll bug @nanli-emory who I think talked me out of the beta version which did just that for additional comments, but maybe this concept would be useful as a user setting for both drawing and rendering? Nan, what do you think?

birm:
Closing for now because stale.

mgautam98:
😕 Okay.

birm:
If you're still interested in this, you're welcome to reopen.

PR #348, opened 2020-04-08
Minor UI changes in upload panel of model app

amritansh22:
Made some minor UI changes in the upload panel of the model app.

- Added focus in input field. The background is light blue in color.

Screenshot (157)

- When there is an error with the input shape or name a red bounding box is drawn around the input field.

Screenshot (158)

PR #360, opened 2020-04-13
Model upload bug

leoarc:
The check duplicate name feature was added in #318 and was fixed in #327 . But the bug was again added in #348 as some of the old code was added back. This PR fixes the bug .

Also earlier the feature worked only after refreshing the table as the ``` modelName ``` was only updated during initialization which meant same name model could be uploaded in the same session. This has also been fixed .

### DIFF
--- a/apps/model/model.js
+++ b/apps/model/model.js
@@ -695,18 +695,7 @@ function uploadModel() {
           document.getElementById('imageSize').style = 'border:2px; border-style: solid; border-color: red;';
           return;
         }
-
-        try {
-          if (checkNameDuplicate(_name.value)) {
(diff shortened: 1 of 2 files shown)


birm:
Ah, this is what happens when we both try to do a fix at the same time. I think that cd00a7b also fixes, but keeps the name duplicate checking; does it work for you?

leoarc:
the ``` namearray ``` is not needed . It is the earlier implementation which doesn't work. The right one was done in #327

birm:
Thanks for catching this. I guess I must've confused a revert for new code, or something like that.

PR #362, opened 2020-04-13
Delete model bug

### LINKED ISSUE #361: Delete model bug
When trying to delete a model the delete dialog box comes up multiple times( no. of models times ) when clicked once . Example : 
bug2

Expected behaviour : The box should come only once when clicked once .

leoarc:
Fixes #361 
It was caused as all the ``` delete model ``` buttons were being assigned the same id . So when one was clicked all of them were triggered. 

This PR also makes the ``` check duplicate name ``` feature compatible with the ``` delete ``` feature . Earlier even after deleting the model, that name couldn't be used again. Now it when the model is deleted the name is also deleted and can be reused .

### DIFF
--- a/apps/model/model.js
+++ b/apps/model/model.js
@@ -792,7 +792,7 @@ async function showInfo() {
             'id=removeModel'+ modelCount+' type="button"><i class="material-icons"'+
             'style="font-size:16px;">delete_forever</i>Remove Model</button>';
             td = row.insertCell();
-            td.innerHTML = '<button class="btn btn-primary btn-xs my-xs-btn" '+
+            td.innerHTML = '<button class="btn-change" '+
             'id=chngClassListBtn'+ modelCount+' type="button"><i class="material-icons"' +
             'style="font-size:16px;">edit</i>  Edit Classes</button>';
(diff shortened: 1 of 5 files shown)


birm:
Would you mind applying the same fix to chngClassListBtn?

leoarc:
My bad , I should have done that already.

apps/model/model.js
```diff
@@ -806,17 +788,18 @@ async function showInfo() {
...
             'style="font-size:16px;">edit</i>  Edit Classes</button>';
-            document.getElementById('removeModel').addEventListener('click', () => {
+            document.getElementById('removeModel'+modelCount).addEventListener('click', () => {
               deleteModel(name);
             });
-            document.getElementById('chngClassListBtn').addEventListener('click', () => {
+            modelCount+=1;
+            document.getElementById('chngClassListBtn'+modelCount).addEventListener('click', () => {
               showNewClassInput(name);
             });
```
birm, line 804: A little issue, but I think swapping these lines will resolve it.
```suggestion
            document.getElementById('chngClassListBtn'+modelCount).addEventListener('click', () => {
              showNewClassInput(name);
            });
            modelCount+=1;
```

PR #365, opened 2020-04-14
Implementation of Rotation Bar

Hemansh31:
This is an **initial implementation** of the rotation bar of OpenSeadragon instance.
caMicroscope (1)
**Current Behaviour**
Currently rotation is enabled only if none of the tools ( measurement tool, draw annotations etc. ) are not selected.
Upon selecting any tool rotation is disabled and image angle is set to 0 degrees again.
This is done because there are issues with the tools when image is rotated.
Maybe we can open separate issues for resolving these problems.

### DIFF
--- a/apps/viewer/uicallbacks.js
+++ b/apps/viewer/uicallbacks.js
@@ -326,6 +326,7 @@ function presetLabelOn(label) {
     label.color;
   // close layers menu
   $UI.layersSideMenu.close();
+  rotationOff();
 }
 
 function presetLabelOff() {
@@ -348,6 +349,7 @@ function presetLabelOff() {
       '';
     $CAMIC.status = null;
   }
+  rotationOn();
 }
 
 function toolsOff() {
@@ -419,6 +421,7 @@ function annotationOn(state, target) {
   const input = $UI.annotOptPanel._form_.querySelector('#name');
   input.focus();
   input.select();
+  rotationOff();
 }
 
 function annotationOff() {
@@ -438,6 +441,7 @@ function annotationOff() {
     toggleOffDrawBtns();
     $CAMIC.status = null;
   }
+  rotationOn();
 }
 
 function toggleOffDrawBtns() {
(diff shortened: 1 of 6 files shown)


cjchirag7:
@Hemansh31 , it appears to be a cool feature !
I would like to suggest some improvements / bug fixes :
- The 'share view' tool  is not storing the rotation currently (Although it stores the zoom value). So, it can be modified such that it also stores the rotation value
- After creating an annotation, even when no tool is selected, the rotation is disabled. Even after hiding all the annotations, the rotation is still disabled. This can be fixed.
rotation-annotation-bug

- In side by side viewer  , the rotation works for only the right image, it doesn't have any effect on the left image.

Hemansh31:
@cjchirag7 Thanks for the suggestions, I will implement it ASAP.

cjchirag7:
@Hemansh31 , also I noticed one thing in the side viewer. If we open side viewer with some initial rotation value, then movement of both the images occur in entirely different frames. 
We expect that the movement should occur in sync, in the view of the user.

Peek 2020-04-14 18-11

Hemansh31:
@cjchirag7 as I have mentioned in the earlier comments, I haven't synced the rotation and it's behaviour with different tools. I will implement them subsequently.

Hemansh31:
@birm Please Review....

birm:
I'm not entirely sure it makes sense to have a viewer function that doesn't work with annotations/heatmaps yet. @nanli-emory your thoughts?

Hemansh31:
@birm I think we can open separate issues for syncing the annotations, heatmaps with the rotation, but this can be a starting point to build upon. Maybe we can develop on a different branch until everything is synced.

nanli-emory:
You have got a point. @birm Could you create a new branch for `rotation`? @Hemansh31 Please develop it on a different branch. We can merge it back when everything is done. :grin::grin::grin:

birm:
Ok, I've created `feature-rotate` for that purpose.

PR #367, opened 2020-04-15
Slide name change functionality  &  direct slide delete in No-Security Mode

akhil-rana:
Slide name can be updated by the user who have the proper permissions.
ezgif-2-43ced679a2b2


akhil-rana:
The delete functionality is now checked using the `user permissions` rather than `userType`, so slide deletion can now be directly done if security is turned off. 
Also, for user who can delete directly, 'delete' text inside the button is replaced by only the icon, like before.

PR #372, opened 2020-04-16
Re-initialize slides table on slide name change and show popup when successful

### LINKED ISSUE #373: Slide name not updating in the table when the update is successful [BUG]
**Describe the bug**
When user updates the name of a slide, the slide name doesn't change until he reloads the page. This can be improved to automatically update the name.

**To Reproduce**
1. Go to the slides page
2. Click on the pencil icon, which appears when you hover over any of the slide names 
3. Click OK on alert showing successful update of file name.
4. Observe that slide name doesn't change
5. Reload the page to observe the change in the slide name.

**Screencast**
bug

**Expected behavior**
The user must not need to reload the page. The name of the slide should be updated automatically when the operation is successful.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Browser Chrome
 - Version 81.0.4044.92 (Official Build) (64-bit)

**Additional context**
This is actually due to the calling of intialise() function outside the promise. Instead, it should be called inside the promise to fix this.

cjchirag7:
##  Description
The following changes have been made :
- Re-initialized slides table on slide name update
- Replaced alert by a popup for UI  improvements, when slide update is successful
- Replaced alert by a popup for UI  improvements, when slide delete is successful
- Made a function for showing popup for a success message and call it at other places, to prevent using the same piece of code again and again
- Made position of popup as 'fixed' for better UI
- Replace alert by inline error message in form-input, when user enters duplicate value of slide name for UI enhancement

## Motivation and Context
- Fixes #373 
- Earlier the function initialise was called outside of the promise, wrongly. This resulted in no initialisation of the slides table occurring, after the name was edited. So, the user was required to reload the page by himself for this.
- Showing a popup for a few seconds, instead of an alert, which forces user to click 'OK'  to make it close, has a better UI 
- Showing alert for wrong form input was again a bad UI practise which motivated me to modify it.

## How Has This Been Tested?
- Tested code in both Chrome and Firefox

## Screencast :
change_slide_name

## Types of changes

## Checklist:

### DIFF
--- a/apps/loader/loader.js
+++ b/apps/loader/loader.js
@@ -190,26 +190,7 @@ function handlePost(filename, slidename, filter, reset) {
             (success) => {
               initialize();
               $('#upload-dialog').modal('hide');
-              // show pop-up message to user
-              let popups = document.getElementById(
-                  'popup-container',
-              );
-              if (popups.childElementCount < 2) {
-                let popupBox = document.createElement('div');
-                popupBox.classList.add('popup-msg', 'slide-in', 'text-success');
-                popupBox.innerHTML = `<i class="fa fa-check-circle" aria-hidden="true"></i>
-                Slide posted sucessfully`;
-                // Add popup box to parent
-                popups.insertBefore(
-                    popupBox,
-                    popups.childNodes[0],
-                );
-                resetUploadForm();
-                setTimeout(function() {
-                  // popups.lastChild.classList.add('slideOut');
-                  popups.removeChild(popups.lastChild);
-                }, 2000);
-              }
+              showSuccessPopup('Slide updated successfully');
               return changeStatus('POST', success.result, reset);
             }, // Handle the success response object
         ).catch(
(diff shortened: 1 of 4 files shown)


Vedant1202:
Hey @cjchirag7, just make sure keeping the position as fixed does'nt make multiple popups overlap over each other. Please check if multiple popups are stacked or visibly distinct in however way you want to place them. Else just use relative positoning.

cjchirag7:
@Vedant1202 , thanks for your suggestion but I don't think that would be an issue, as the popup will disappear itself after an interval of 2 seconds. I tried using 'relative', but it was not working as expected.

Vedant1202:
@cjchirag7 Well okay, just trying to handle an edge case.

If you're facing issues with placement, you could try having a container with fixed position and a higher z-index. Then put up the popups with position relative, inside the container. Upto you though.

cjchirag7:
@Vedant1202 , yeah works ! thanks again :smiley:

PR #375, opened 2020-04-16
[Enhancement] Model upload and delete UI

mgautam98:
## Description
After uploading model the form will close and a popup will show that the model is uploaded successfully. Previously, user was directed to manually refresh.

After deleting a model a popup for confirmation is shown to the user instead of alert("Deleted"). Also initialized the UI after deletion. 

Deleted popup.css from apps/viewer it is not used anywhere instead css/popup.css is used

## Motivation and Context
- To improve user experience  

## Types of changes

## Checklist:


apps/model/model.js
```diff
@@ -743,9 +754,18 @@ async function deleteModel(name) {
       alert(err);
     } finally {
       if (status) {
-        alert('Deleted', name);
-        showInfo();
-        modelName.splice(modelName.indexOf(name.split('/').pop().split('_').splice(2).join('_').slice(0, -3)), 1);
```
leoarc, line 748: Hey @mgautam98 . I had added this line to remove the model name from ``` modelName ``` array so that that name could be reused once that model has been deleted. Is it not required ?

apps/model/model.js
```diff
@@ -730,7 +740,8 @@ function uploadModel() {
 }
 
 async function deleteModel(name) {
-  if (confirm('Are you sure you want to delete this model?')) {
+  modelName = name.split('/').pop().split('_').splice(2).join('_').slice(0, -3);
```
leoarc, line 743: ```  modelName ``` is the global array which stores the names of the model for the ``` checkDuplicateNameFunction ``` . I think you should use a different variable name in line 743 or else the array is reinitialized wrongly every time ``` deleteModel ``` is called . Am  I missing something @mgautam98 ?

apps/model/model.js
mgautam98: This line is not needed anymore. After deletion I am calling initUIComponent which will fill the modelName array again.

apps/model/model.js
mgautam98: Yes, you are right. A different name should have been used I missed it. It will not cause any problem because after deleting the modelName will be reinitialized so no a problem. But It is confusing and I will correct it in next PR or you can do it too if you want in you next PR.

apps/model/model.js
leoarc: Oh sorry for the confusion. I missed that . Thanks for clearing that for me!

PR #382, opened 2020-04-17
[Enhancement] Popup display in model and segment app

amritansh22:
## Description
In model app:

- After changing class list, instead of an alert a popup is shown.
- UI of class change button

In segment app

- After uploading model, a popup is shown.
- After deleting model, instead of an alert a popup is shown.

## Motivation and Context
Improves User Interface

## How Has This Been Tested?
Tested in Chrome and Firefox

## Types of changes

## Checklist:

### DIFF
--- a/apps/segment/segment.js
+++ b/apps/segment/segment.js
@@ -1191,7 +1191,7 @@ async function showInfo() {
             document.getElementById('removeModel'+modelCount).addEventListener('click', () => {
               deleteModel(name);
             });
-           modelCount+=1;
+            modelCount+=1;
           };
         }
       }
--- a/apps/segment/segment.js
+++ b/apps/segment/segment.js
@@ -1133,6 +1133,7 @@ async function deleteModel(name) {
       alert(err);
     } finally {
       if (status) {
+        modelName.splice(modelName.indexOf(name.split('_').splice(1).join('_').slice(0, -3)), 1);
         let popups = document.getElementById('popup-container');
         if (popups.childElementCount < 2) {
           let popupBox = document.createElement('div');
--- a/apps/segment/segment.js
+++ b/apps/segment/segment.js
@@ -1158,7 +1158,7 @@ async function showInfo() {
   var table = document.querySelector('#mdata');
   var tx = db.transaction('models_store', 'readonly');
   var store = tx.objectStore('models_store');
-
+  var modelCount=0;
   empty(table);
   // Update table data
   (function(callback) {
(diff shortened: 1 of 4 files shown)


apps/segment/segment.js
```diff
@@ -1166,13 +1183,12 @@ async function showInfo() {
             td = row.insertCell();
             td.innerHTML = date;
             td = row.insertCell();
-            td.innerHTML = '<button class="btn-del" '+
-            'id=removeModel'+modelCount+' type="button"><i class="material-icons"'+
+            td.innerHTML = '<button class="btn btn-primary btn-xs my-xs-btn" '+
+            'id="removeModel" type="button"><i class="material-icons"'+
```
leoarc, line 1187: This will bring back #361 . I think you don't have the latest ``` develop ``` branch . These changes were made in #362  and I think they shouldn't be changed back.

apps/segment/segment.js
```diff
@@ -1169,10 +1186,9 @@ async function showInfo() {
             td.innerHTML = '<button class="btn-del" '+
             'id=removeModel'+modelCount+' type="button"><i class="material-icons"'+
             'style="font-size:16px;">delete_forever</i>Remove Model</button>';
-            document.getElementById('removeModel'+modelCount).addEventListener('click', () => {
+            document.getElementById('removeModel').addEventListener('click', () => {
               deleteModel(name);
             });
-            modelCount+=1;
```
leoarc, line 1175: These changes should be reverted too  I think .

leoarc, line 1175: These changes need to be reverted too I think .

apps/segment/segment.js
```diff
@@ -1172,7 +1191,7 @@ async function showInfo() {
             document.getElementById('removeModel'+modelCount).addEventListener('click', () => {
               deleteModel(name);
             });
-            modelCount+=1;
+           modelCount+=1;
```
birm, line 1194: this should fix the style build
```suggestion
            modelCount+=1;
```

apps/segment/segment.js
amritansh22: Thanks for pointing out. I have fixed it.

apps/segment/segment.js
amritansh22: Ya thanks.  I have done it.

apps/segment/segment.js
amritansh22: Thanks for pointing out. I have fixed it.

birm:
I'm still getting an alert on model.html on changing classes. Have I missed/misunderstood something?

amritansh22:
@birm I had missed that. Sorry about that.

PR #383, opened 2020-04-17
[Enhancement] replace alerts to toasts in viewer

mgautam98:
## Description
Toast for no heatmap data instead of alert
Toast for no markup on annotation instead of alert
Documented functions inside viewer uicallbacks.js

- segment.js and model.js
Changed name while deleting model to avoid confusion which was created in #382 and #375  

## Motivation and Context
- Documenting code is good for fast development. 
- Replacing alerts with toasts to improve UX

## Types of changes

## Checklist:

### DIFF
--- a/apps/model/model.js
+++ b/apps/model/model.js
@@ -740,8 +740,8 @@ function uploadModel() {
 }
 
 async function deleteModel(name) {
-  modelName = name.split('/').pop().split('_').splice(2).join('_').slice(0, -3);
-  if (confirm('Are you sure you want to delete ' + modelName + ' model?')) {
+  DeletedmodelName = name.split('/').pop().split('_').splice(2).join('_').slice(0, -3);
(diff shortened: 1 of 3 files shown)


apps/model/model.js
```diff
@@ -740,8 +740,8 @@ function uploadModel() {
 }
 
 async function deleteModel(name) {
-  modelName = name.split('/').pop().split('_').splice(2).join('_').slice(0, -3);
-  if (confirm('Are you sure you want to delete ' + modelName + ' model?')) {
+  DeletedmodelName = name.split('/').pop().split('_').splice(2).join('_').slice(0, -3);
```
birm, line 743: sorry to nitpick, but can we do camelCase?
```suggestion
  deletedModelName = name.split('/').pop().split('_').splice(2).join('_').slice(0, -3);
```
(Thanks for fixing this!)

mgautam98: Thanks I will change all the instances of DeletedmodelName

PR #391, opened 2020-04-21
Add filter UI in slide table and information dashboard [Feature]

### LINKED ISSUE #386: Table UI Around Filters
With the new backend, we have a filters concept (that is, users can be given a list of filters, and cannot see slides not in that list.)

Conceptually, this task would involve giving a user some way to select a filter, and only see slides which have that filter. (e.g. Alice has access to "Secret" and "Super Secret" via their user filter, and wants to only see "Secret")

cjchirag7:
##  Description
Following changes have been made
- Add filter UI in slide table and information dashboard
- Validate filter when user uploads a slide.
- Simplify code for searching in both table.html and info.html
- Made sure that searching, sorting and filtering, all work together without any collision.
- The filters are shown only in security mode.

## Motivation and Context
- Fixes #386 

## How Has This Been Tested?
- Tested both desktop and mobile view in Chrome and Firefox by once login in as 'Admin', once as 'Editor' with some filters and once by setting `DISABLE_SEC` and `ALLOW_PUBLIC` as true

## Screencasts -

1. Filters UI in table.html (Similar exists in Info.html)
filter-ui

2. Validation for filter during Slide Upload
filter-validation

## Types of changes

## Checklist:

### DIFF
--- a/apps/Info.html
+++ b/apps/Info.html
@@ -85,6 +85,10 @@
 
 	<div class="container">
 	<h3 class="text-center h3 mb-2" style="margin-top:8px;margin-bottom:10px;">Information Dashboard</h3>
+	<div class="row mt-2" id="filters-heading">
+	</div>
+	<div class="row mb-2 ml-1" id="filters-check">
+	</div>
 	<div class="form-group has-search">
 		<span class="fa fa-search form-control-feedback"></span>
 		<input id="search-table" type="text" class="form-control" placeholder="Search">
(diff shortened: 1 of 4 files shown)


birm:
As of now, we don't support having an array of filters on an object -- https://github.com/camicroscope/Caracal/blob/master/handlers/filterFunction.js
You could add support for this if you want. (presumably we'd allow if the user filter array and the object filter array have anything in common?)

cjchirag7:
@birm  sorry, I didn't notice that earlier. But, I have added support for an array of filters in #camicroscope/Caracal/pull/28

PR #399, opened 2020-05-28
Safe html in table

birm:
* add linting to table.js
* try to mitigate unsafe html rendering which is all over table.js

While using document.createElement and related would be best, this would have taken to long, so after I gave up on that, I just used the sanitize function made earlier.


birm:
looked ok with test. Test again before next release.

PR #409, opened 2020-06-10
Initial Design/Workflow for Development Workbench + Initial implementation of Spritesheet generator.

akhil-rana:
## This is the initial implementation for the design and workflow for the Development Workbench. 
~~This works with [Caracal #38](https://github.com/camicroscope/Caracal/pull/38)~~
This works with [SlideLoader #32](https://github.com/camicroscope/SlideLoader/pull/32)

## Description
1. Stepper interface is used so that it's easy for the user to navigate sequentially.
2. Initial implementation of dataset selection has been shown where user can either choose a pre-made (formatted i.e spritesheet & labels) dataset or can create an own dataset from labelling files from caMircoscope. 
3. When creating own dataset user can choose which label images to use. No of images under every label is shown and user can select which of them to keep.
4. The user will be returned with a zip file containing the spritesheet and binary labelling according to the selection.
5. The corresponding zip file can then be browsed to proceed to the next section. 

## Motivation and Context
- This is a first step in training own ML models from the usable dataset returned by the spritesheet generator which can be directly used into any TF.js project. 

## How Has This Been Tested?
- Briefly tested on Firefox (78.0), Chrome (84.0)  on Windows 10 and Linux (arch)

## Screenshots:
input

## Types of changes

## Checklist:

### DIFF
--- a/apps/dev-workbench/workbench.js
+++ b/apps/dev-workbench/workbench.js
@@ -1,18 +1,3 @@
-// let JSZip = require("jszip");
-// let zip = new JSZip();
-// zip.generateAsync({type: 'blob'})
-//     .then(function(blob) {
-//       saveAs(blob, 'hello.zip');
-//     });
-// function saveAs(theBlob, fileName) {
-//   // A Blob() is almost a File() - it's just missing the two properties below which we will add
-//   theBlob.lastModifiedDate = new Date();
-//   theBlob.name = fileName;
-//   return theBlob;
-// }
-
-// $('#labelsInput').unbind('change');
-
 function dataSelect() {
   $('#stepper').hide(300);
   $('#headContent').hide(300);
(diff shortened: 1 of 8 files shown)


apps/dev-workbench/workbench.js
```diff
@@ -0,0 +1,224 @@
...
+      <div class="custom-control custom-checkbox">
+        <input
+          type="checkbox"
+          class="custom-control-input"
+          id="labelCheck` +
+        i +
+        `" checked /> <label class="custom-control-label" for="labelCheck` +
+        i +
+        `" >` +
+        data.labels[i] +
```
birm, line 153: Can you either sanitize or use innerText to prevent xss for things like this?

akhil-rana: Yes, I should do that. Thanks for pointing out.

akhil-rana: I sanitised the user inputs accordingly. I think that should be safer?

birm: I'm not sure which would be safer, but as far as I'm concerned, both are fine. :smile:

akhil-rana:
> Do you think it makes sense to add the workbench link to the table.html bar too?

I thought that would be easily accessible. Though I'm not sure about the placement of the workbench link. 
We can also add it on just the landing page making a new tile with the others.

akhil-rana:
> Looks good; though I do still think adding workbench to table.html would be nice :)

Where do you mean to add it? In the nav bar? I've already added it in table.html nav bar.

birm:
> Where do you mean to add it? In the nav bar? I've already added it in table.html nav bar.

My bad, in landing.html

akhil-rana:
> > Where do you mean to add it? In the nav bar? I've already added it in table.html nav bar.
> 
> My bad, in landing.html

Yeah I was thinking about that. Should've asked you. I will add it. Thanks.

akhil-rana:
- Added support for custom dataset.
- Added support for user-customisable normalised resolution for the images in the dataset. 

ezgif-6-fa19726e0887

birm:
Awesome, no need to hide your work in the menus :smile: 
Just confirm when you feel comfortable to merge.

akhil-rana:
I think we can merge now. Any suggestions are welcomed. I've started with the trainer now so if anything else is needed to be changed here I'll push it along with that.

PR #417, opened 2020-06-30
Use backend to download and update download folder structure

leoarc:
## Description
- Added feature to give user choice to use  SlideLoader to download the patches
- Updated the download folder structure so that now each class is downloaded in a separate folder with the class name

## Motivation and Context
- To make roi feature usable with large slide 
- In many places , datasets divided into folders according to their classes is used directly 

## How Has This Been Tested?
Ubuntu 20.04
Firefox 

## Types of changes

## Checklist:

This depends on [SlideLoader #34 ](https://github.com/camicroscope/SlideLoader/pull/34)

### DIFF
--- a/apps/model/model.css
+++ b/apps/model/model.css
@@ -511,3 +511,63 @@ body{
   from {bottom: 30px; opacity: 1;}
   to {bottom: 0; opacity: 0;}
 }
+
+.switch1 {
+    position: relative;
+    display: inline-block;
+    width: 100px;
+    height: 26px;
+    border: 1px solid black;
+}
+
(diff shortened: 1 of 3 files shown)


leoarc:
As discussed with  @Insiyaa now instead of only taking into account the max accurate class , it tries to consider all classes to find matches and then chooses the one with most accuracy within them . So now if A = 0.6 , B= 0.3 , C=0.2 and user selects B,C,0.1 it will give B,0.3 ( among B and C the one with more accuracy is chosen)

Insiyaa:
Looks good to me!

PR #434, opened 2020-08-06
warn just once about self-intersecting polygons

birm:
Re "FAST" Meeting; use localstorage to check if we've warned the user about polygons before. Still do a console info in case someone later wonders why the area/perimeter fails.


birm:
Quick test worked for me.

PR #435, opened 2020-08-06
Use vw for toolbar button sizes

### LINKED ISSUE #431: Scaling of UI Elements
Ui Elements, especially the toolbar, is too small on high res screens

birm:
closes #431
Draft, since it's super minimal and I don't like using !important.
The below was taken on a simulated 3500 px wide monitor
Screenshot from 2020-08-06 00-56-28

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -251,7 +251,7 @@ background: #71daec !important;
 opacity: 0.3 !important;
 }
 
-@media (max-width: 1200px) {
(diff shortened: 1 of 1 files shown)


birm:
Should we extend it to apply to other ui elements too?

nanli-emory:
I think this is a quick fix. we could start a new issue for all UI elements and start to work on it later. Sounds good?

PR #438, opened 2020-08-13
Workbench

birm:
Placeholder PR, for now.

We'll want to bring in the GSOC code when it's ready.


birm:
@akhil-rana  let me know when you're ready for us to merge this PR :)

apps/landing/landing.html
```diff
@@ -50,6 +50,12 @@
             <li class="nav-item  link" style="font-family: sans-serif;">
                 <a class="nav-link" href="../table.html"> <i class="fas fa-list-ul"></i>  Slides</a>
             </li>
+            <li class="nav-item link">
```
akhil-rana, line 53: We're now having the `info` tab in NavBar in landing.html, table.html, info.html which was removed from navbar in #376

akhil-rana:
And also I'm not sure if the footer should be like this when there are less number of slides.

DeepinScreenshot_select-area_20200830030504

nanli-emory:
> And also I'm not sure if the footer should be like this when there are less number of slides.
> 
> DeepinScreenshot_select-area_20200830030504

when the screen width is small, the footer becomes this way. It should be the data table things.

akhil-rana:
Do you think we should remove the `info` option from the navbar? Since it is now present at multiple places.

nanli-emory:
> Since



> Do you think we should remove the `info` option from the navbar? Since it is now present at multiple places.

Don't worry about the `info` option. We can keep it ~~~~ 💯

akhil-rana:
Okay. Great. 
Thank you.

akhil-rana:
I think we should be ready to merge into develop now.

Issue #413, opened 2020-06-24
Improved In-Code Documentation


birm:
Some things are pretty difficult to read, and could likely use a bit of help.
Thus, describe at a high/medium level (as appropriate) as code comments.
Use judgement to determine if something is already clear.
(that is the following is probably not necessary)
```js
// performs normalization on an image
normalizeImage(img)
```
We'll probably break this into subtasks, depending on what PRs come in.

Saurav-Majumder:
Can I help through it

anumi1999:
Is this issue still open?

birm:
Yes, thanks for your interest. I'd recommend doing some small sections first so we can review before you put too much time into it.

anumi1999:
Made some changes in the documentation please have a look and review

AkhileshAdithya:
Is this issue still open? Are contributions welcome?

garvpatidar04:
is this still open? Can I contribute to it?

ashutosh887:
I would like to work on this Issue. @birm

AdarshRawat1:
May I inquire whether this opportunity is presently available for contribution? If so, I would appreciate the chance to make a contribution.

birm:
We welcome pull requests on open issues from anyone!

Znoy108x:
I would like to contribute in this !

NikhilKnox:
is this open? Can i contribute!

Sankalp-sa:
I would like to contribute int this

Sam-1806:
Is this open? Can I contribute?

Shubhisingh921:
I would like to contribute in this as i am good at documentation.

AmanGuptaCoder:
I would like to contribute in documentation.

affusul:
i would like to contribute to this,if it is open.

birm:
I'm going to close this issue, not because it is done, but because it can never really be "finished". 
I will still accept pull requests which meaningfully improve documentation.

niikkhilsharma:
Hey, Birm
I understand your concern but I would like to give it a try. I am good at managing and writing documentation. I have written couple of documentation for several projects and I believe that I can help at this. I hope you will give me a chance. 
Thanking You

PR #478, opened 2021-01-23
Improved the Documentation  #413

anumi1999:
## Description
**Describe your changes in detail
The code documentation for google login is improved. Meaning of each variable is defined clearly.
## Types of changes

## Checklist:
- [ ] My code follows the code style of this project.


birm (approved):
Interesting choice to document first, but looks reasonable to me. I'm not sure that some of the callback descriptors are necessary, but should be better than not having them. Thanks for taking the time.

PR #479, opened 2021-01-25
Update README Documentation

sohamsshah:
## Description
The reference to setup Cloud Upload APIs for enabling the 'Upload from Drive' and similar features for caMicroscope that is available in the file - docs/cloudUploadAPIs.md is missing in the README.md

In this PR, it has been appended to the README.md file.

## Types of changes
**What types of changes does your code introduce? Put an `x` in all the boxes that apply:
- [] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
**Go over all the following points, and put an `x` in all the boxes that apply.
**(If you're unsure about any of these, don't hesitate to ask. We're here to help!)
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

PR #486, opened 2021-02-09
add Dropbox upload feature

sohamsshah:
## Description
This PR adds the Dropbox Upload Feature to the project. It provides flexibility to choose the files from the Dropbox platform too.
Along with that I have updated the `README.md` file that solves #479 as it was linked with cloud uploads. Documented the instructions on how to setup and steps in the `cloudUploadAPIs.md` file too.

## How Has This Been Tested?
- As suggested by @akhil-rana I have tested the feature by checking the upload of large files from Dropbox. The files that are used for testing are taken from [Open Slides](http://openslide.cs.cmu.edu/) of different formats and sizes.
- The feature is working and tested.

## Video Demo

https://user-images.githubusercontent.com/47717492/107621272-fd37c580-6c7b-11eb-9a58-a690dcd3610d.mp4

File uploaded - CMU-1.svs
Size - 169 MB

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

Please review this and let me know any sort of changes or suggestions are needed to be implemented to get it merged. Would revert to the comments/reviews actively :) Thank you!

### DIFF
--- a/apps/loader/chunked_upload.js
+++ b/apps/loader/chunked_upload.js
@@ -402,13 +402,13 @@ function continueGoogleDriveUpload(token) {
 }
 
 // to upload a file from Dropbox
-function dropboxChooserStart(){
-  //refer docs: https://www.dropbox.com/developers/chooser
+function dropboxChooserStart() {
+  // refer docs: https://www.dropbox.com/developers/chooser
   options = {
 
     // Required. Called when a user selects an item in the Chooser.
     success: function(files) {
-        handleUpload(files);
+      handleUpload(files);
     },
 
     // Called when the user closes the dialog without selecting a file
@@ -419,7 +419,7 @@ function dropboxChooserStart(){
 
     // Optional. "preview" (default) is a preview link to the document for sharing,
     // "direct" is an expiring link to download the contents of the file.
-    linkType: "direct", // or "direct"
+    linkType: 'direct', // or "direct"
 
     // Optional. A value of false (default) limits selection to a single file, while
     // true enables multiple file selection.
@@ -427,8 +427,8 @@ function dropboxChooserStart(){
 
     // Optional. This is a list of file extensions. If specified, the user will
     // only be able to select files with these extensions. You may also specify
-    // file types, such as "video" or "images" in the list. 
-    //extensions: ['.pdf', '.doc', '.docx'],
+    // file types, such as "video" or "images" in the list.
+    // extensions: ['.pdf', '.doc', '.docx'],
 
     // Optional. A value of false (default) limits selection to files,
     // while true allows the user to select both folders and files.
@@ -436,10 +436,9 @@ function dropboxChooserStart(){
     folderselect: false, // or true
 
     // A limit on the size of each file that may be selected, in bytes.
-    //sizeLimit: 1024, // or any positive number
-};
+    // sizeLimit: 1024, // or any positive number
+  };
 
-//triggers dropbox chooser
-Dropbox.choose(options);
-
+  // triggers dropbox chooser
+  Dropbox.choose(options);
 }
--- a/apps/loader/chunked_upload.js
+++ b/apps/loader/chunked_upload.js
@@ -441,5 +441,7 @@ function dropboxChooserStart() {
   };
 
   // triggers dropbox chooser
+  
+  Dropbox.appKey = dropboxConfig.APP_KEY;
   Dropbox.choose(options);
 }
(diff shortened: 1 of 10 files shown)


apps/loader/chunked_upload.js
```diff
@@ -400,3 +400,46 @@ function continueGoogleDriveUpload(token) {
...
+
+    },
+
+    // Optional. "preview" (default) is a preview link to the document for sharing,
+    // "direct" is an expiring link to download the contents of the file.
+    linkType: "direct", // or "direct"
+
+    // Optional. A value of false (default) limits selection to a single file, while
+    // true enables multiple file selection.
+    multiselect: true, // or false
```
akhil-rana, line 426: I'm not sure if you should leave this as true.
Since multi file upload is not allowed from the normal upload flow. We have batch loader for that.
@birm What do you think?

sohamsshah: Yes, makes sense. Thanks for pointing it out!

sohamsshah: *disable multiselect in the following commit

akhil-rana:
First of all thanks for the giving your time into this. @sohamsshah 

So since this is all taking place in the browser. I suppose user's data will be used to download the file from dropbox and same amount of data again to upload (in case of remote hosting)

The main advantage of cloud upload feature is to give user the functionality of uploading to caMicroscope servers directly from a cloud service without using any data from user's side. Like we did in Google drive.
Though this is a not a useless addition, but it gives the same output as if user downloads the file himself and then uploading it.

Please correct me if I'm wrong anywhere.

@birm  Please give your thoughts regarding this.

birm:
Having users add api keys to table.html probably isn't an ideal way to deploy this (we'd have to mount in a custom version like we do with login.html). I'd recommend exposing a new static file with this conf (that is a JSON with any api keys) if you want to keep this entirely frontend. Alternatively, you could follow @akhil-rana's recommendation and move getting the slides to the loader service. I have no objection to either, as both methods are an improvement and new feature. :)

sohamsshah:
Sounds good! Was curious that in case if we are going to add such more API secret keys, can we add a `.env` file to store all the keys there?

sohamsshah:
dropbox-feature

Enhancement: Add extension checking from the Dropbox Upload window itself.
How it helps: Users Experience improves as only `accepted` formats could be selected from the Dropbox Chooser. Cross checked with the Chooser API docs.

Let me know your suggestions on this!

Also fixed linting errors.
In `layersviewer.js` on running `npm run lint-fix` throws error of bad function naming. Would it be fine if I fix in this PR or create separate one fixing this?

Thank you for the review :)

akhil-rana:
> Sounds good! Was curious that in case if we are going to add such more API secret keys, can we add a `.env` file to store all the keys there?

I'm not sure about how env will work in docker. 
And is this really a "secret key"?
Since this is a frontend API, we can't hide the key anyway. They're not meant to be hidden as much as I know. But they need to be restricted through the API provider settings I think.
And even if you want to implement this in the backendin future, you still have to use a picker in the frontend which will require keys. Simply put, frontend api keys are not secret.

So long story short,
Though I'm not sure but it would be really simple if you just create a json config like @birm suggested.
Not to hide the key. But just to make it more straight forward for the end user. Since going through a long html file and editing it is a not that user friendly.

sohamsshah:
Thank you so much @akhil-rana for the clarification, appreciate it. On it! 🎯

sohamsshah:
I have:
- added config file
- It was not possible to edit the attributes of the script tag directly.
- After a bit research on official docs and some console logs, the `Dropbox` object actually handles the URL params of the opened window.
- So `Dropbox.appKey` can be assigned to the APP_KEY from the `config.json`

Let me know how does this sound!

apps/loader/config.js
```diff
@@ -0,0 +1,3 @@
+const dropboxConfig = {
+  'APP_KEY': '6yibvsbuan5leo7',
```
birm, line 2: If this is a valuable API key, you can rebase it out.

sohamsshah: Thanks for the review. Resolved that. The `cloudUploadAPIs.md` covers more upon on how to get the APP_KEY configured. and it has been updated in the following commits.

birm:
@sohamsshah do you need any help with these merge conflicts?

birm:
superseded by #498

Issue #481, opened 2021-01-25
[IMPROVEMENT]: Tour resumes from same place when it was left earlier when restarted


sohamsshah:
**Is your feature request related to a problem? Please describe.**
In the viewer section, the user has an option to take a tour of the application. After the tour is over, the user clicks `end tour` button. 

Consider the case when the User has taken complete tour and clicks on end tour. Then the tour can never start from beginning and always start from the end.

**Describe the solution you'd like**
According to the tour websites I have came across most of them are of 'Always start from beginning' types. So the code file can be edited accordingly.

**Describe alternatives you've considered**
We can have 2 alternatives:
Start the tour always from beginning || (Ask the user to choose which option they want - Resume where they left || Start from beginning).

Also, the `tut.js` doesn't use `backticks` ES6 feature and if there are other refactors needed, will address them.

Both can be implemented with some code changes, let me know which sounds better. Will link PR accordingly.

Thank you.

sohamsshah:
Closing this issue; since this is resolved :)

PR #487, opened 2021-02-09
improve tour resume UX

sohamsshah:
## Description
Solves #481 
It is a UX improvement in tour.
What this change does?
- The tour that is provided for ease for users to understand if left in between, continues from where the user left i.e. default.
- But generally by UX perspective, the tour must start from the beginning because of the following problems:
  * If the tour is left at the end, it resumes from the end and doesn't again point to start.
  * If the user closes the window and when opens again, the local storage has stored the value of last left tour element. And thus will resume from there itself; which is unexpected for user as User would expect it to be from beginning only.

## How Has This Been Tested?
- It changes `tut.js` file. It has been tested for its working on local set up.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] My code follows the code style of this project.

Please let me know how does this sounds. Would love to know your suggestions or improvements on this!

### DIFF
--- a/apps/viewer/tut.js
+++ b/apps/viewer/tut.js
@@ -138,9 +138,9 @@ var tour = new Tour({
   '</div><div style=\'width:100%;display: flex;margin-bottom: 5px;\'>'+
   '<button class=\'btn btn-default\' data-role=\'end\' style=\'margin:auto;\'>End tour</button></div></div>',
   backdrop: true,
-  //resets tour always to resume always from the beginning
-  onEnd: function reset(){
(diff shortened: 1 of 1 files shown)


birm (changes requested):
Can you fix the lint job? `npm run lint-fix` should probably do it.

sohamsshah:
Hi, anything I shall do to get the PR merged?

birm (approved):
Thanks for the fix, and the improvement!

PR #500, opened 2021-03-08
Add Image enhancements

r7rohan:
## Image Enhancement
A new feature, This could be used for medical-image enhancements and processing

### Description
- Histogram Equalization (CLAHE) : used for auto-setting brightness and contrast
- Edge detection and sharpness filter
- Apply filter to a particular region (useful for auto-align etc)

## Context
In this PR, it is applied only to the window-canvas of the viewer page, but later on it would be better if the whole can be processed and saved and toggled (independent of zoom and movement)

## How Has This Been Tested?
- It works well
- Edge detection for the canvas takes 3 secs (at max), others are instantaneous

## Demo:
[Demo video](https://drive.google.com/file/d/1Kff8Bbf_a2ODavh2RrRBOb0EvIYjf_vW/view?usp=sharing)

## Types of changes

## Checklist:


nanli-emory (approved):

birm:
This is a cool concept, but I think this needs either a UI or functionality change to avoid confusing people. When I zoom/pan on an image, it undoes the enhancement. So, if we could either untoggle the button in such an event, or keep it active until deactivated, I think this would be ready to merge.
(Or, let us know if you have an other thought of how this tool should work)

r7rohan:
There could be 2 ways of adding, 
- Either user can enhance the current window (as in this PR) 
- Or the whole slide(/part) is selected and processed and saved. 
There could also be another method for reapplying these enhancements on every change of frame but that has a bit of lag

core/openseadragon/openseadragon.js
```diff
@@ -10172,6 +10172,7 @@ function updateOnce( viewer ) {
          * @property {OpenSeadragon.Viewer} eventSource - A reference to the Viewer which raised this event.
          * @property {?Object} userData - Arbitrary subscriber-defined object.
          */
+        unenhance();
```
birm, line 10175: Openseadragon is used more broadly than just viewer, so this breaks all other apps.

It may be more useful to use an event handler? The [openseadragon docs](https://openseadragon.github.io/docs/OpenSeadragon.EventSource.html) may be helpful for that.

birm (changes requested):

birm (approved):
Thank you for following all of the comments. This is a really cool addition!!

PR #501, opened 2021-03-15
Regex validation for email during signup process

Eddges:
## Description
** Regex validation for email in the signup form. The program throws an alert and doesn't proceed to registration.

## Motivation and Context
** During user registration, it is possible currently to sign up with an invalid email. I felt adding an email validation with a tried and  tested regular expression so the form doesn't accept invalid email might be a good addition.

## How Has This Been Tested?
** The regex is based on RFC822 standard, so it covers the entire spectrum of email addresses. I have used the regex from w3resources. So it is expected to work right.
I tested the same code with camicroscope. It works for all the cases. I have used the same code in my earlier projects and it has never caused an issue.

## Types of changes

## Checklist:


birm (approved):
Thank you, this should hopefully help avoid some confusion

PR #508, opened 2021-03-22
Changed the UI of the Index Page

sarthak1122:
## Description
**Describe your changes in detail

## Motivation and Context
- If it fixes an open issue, please link to the issue here.

## How Has This Been Tested?
- Include details of your testing environment, tests ran to see how
- your change affects other areas of the code, etc.

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -14,7 +14,7 @@ body {
 
 body {
   color: hsl(0, 0%, 100%);
-  background-color: white;
+  background-color: darkseagreen;
   background-size: 2rem 2rem;
   background-position: 0 0, 5rem 0, 5rem -5rem, 0px 5rem;
   background-color: rgba(220, 220, 220, 1);
@@ -76,7 +76,7 @@ body {
 /* label */
 .item_content label {
   /*font-size: 1.5rem;*/
-  border: none;
+  border: royalblue;
 }
 
 /* for button */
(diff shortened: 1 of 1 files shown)


birm:
Hello. Can you describe what changes you're intending to make here and why?

css/style.css
```diff
@@ -14,7 +14,7 @@ body {
 
 body {
   color: hsl(0, 0%, 100%);
-  background-color: white;
+  background-color: darkseagreen;
```
sohamsshah, line 17: Since you are updating the color, wouldn't it be good if you take the colors out to the root and define the colors for the page there itself?

Something of this sort 
```
:root{
    --primary-color: "primary-color"
    --secondary-color: "secondary-color"
// similar kind of thing
```
[This](https://www.w3schools.com/css/css3_variables.asp) can be useful to take a quick glance.

So by this we can refactor the color scheme to the root and becomes easily accessible across the code. 
Btw please confirm that on changing this style, other pages also reflect the similar styling to avoid inconsistency.

Thank you.

sohamsshah (changes requested):

css/style.css
sarthak1122: @sohamsshah  It would be a good idea to take the color attribute out of the root but it is one of the way to style our page . 
The above code would render the home page so Changes made in background color would reflect on the home page only !

You are Welcome to ask more questions :)

sarthak1122:
@brim the changes that I made is mainly in styling of the page  including some transition effects and adding some background color and border to the above code !

birm:
Do you have a screenshot?

PR #511, opened 2021-04-01
Feature for additional links in toolbar

### LINKED ISSUE #468: Formalized Way to add Link Buttons
Very frequently, I'm asked to add links to forms to deployments. While this isn't difficult, it seems to be a well-used enough feature to add a formal way here and in camicroscope/Distro to add these link buttons. (e.g. a list of the button icon name and url?)

Vedant1202:
Feature for additional links in toolbar of viewer.

## Summary

This change provides the ability to put additional links in the toolbar for the slide viewer. This PR should solve #468 . This feature also needs a JSON file which is described in [Distro PR-161](https://github.com/camicroscope/Distro/pull/161).

InkedVirtualBox_Ubuntu-18-GSOC_01_04_2021_17_45_12_crop_LI
The new links are appended as shown with the green marker.
The links are customizable with the following parameters:
1. The URL
2. The name to be displayed
3. The icon to be displayed (material icons)
4. A boolean to choose if the link is to be opened in new tab, or in current tab.

## Motivation

Yes it fixes #468 

## Testing

This has been tested on local system running Ubuntu 18.04, docker-compose version 1.28.6, and Mozilla firefox browser.

### DIFF
--- a/apps/viewer/init.js
+++ b/apps/viewer/init.js
@@ -570,28 +570,32 @@ async function initUIcomponents() {
   var additionalLinksFetchResponse = await fetch('http://localhost:4010/additional_links.json', {headers: headers});
   // Handle error
   if (!additionalLinksFetchResponse.ok) {
-    var message = `Error ${additionalLinksFetchResponse.status}: File not found or file not in correct format`;
-    alert(message);
-    throw new Error(message);
+    var message = `Error ${additionalLinksFetchResponse.status}: File containing JSON data for additional links not found`;
+    console.error(message);
+  } else {
+    try {
+      var additionalLinks = await additionalLinksFetchResponse.json();
+  
+      additionalLinks.forEach(function(additionalLink) {
+        var openInNewTab = additionalLink.openInNewTab === false ? false : true;
+        var url = additionalLink.url;
+    
+        subToolsOpt.push({
+          name: additionalLink.displayName,
+          icon: additionalLink.icon ? additionalLink.icon : 'link',
+          title: additionalLink.displayName,
+          value: additionalLink.displayName,
+          type: 'btn',
+          callback: function() {
+            additionalLinksHandler(url, openInNewTab);
+          },
+        });
+      });
+    } catch (error) {
+      console.error(error);
+    }
   }
 
-  var additionalLinks = await additionalLinksFetchResponse.json();
-
-  additionalLinks.forEach(function(additionalLink) {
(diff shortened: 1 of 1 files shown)


birm (changes requested):
Please catch the errors in the promise chain such that a malformed/missing additional_links file fails gracefully

Vedant1202:
@birm I've added the error handler. Let me know if that works fine.

apps/viewer/init.js
```diff
@@ -553,6 +553,46 @@ async function initUIcomponents() {
...
+
+  var additionalLinksFetchResponse = await fetch('http://localhost:4010/additional_links.json', {headers: headers});
+  // Handle error
+  if (!additionalLinksFetchResponse.ok) {
+    var message = `Error ${additionalLinksFetchResponse.status}: File not found or file not in correct format`;
+    alert(message);
+    throw new Error(message);
+  }
+
+  var additionalLinks = await additionalLinksFetchResponse.json();
```
birm, line 578: A malformed json still fails here.

apps/viewer/init.js
```diff
@@ -553,6 +553,46 @@ async function initUIcomponents() {
...
+      window.location.href = url;
+    }
+  }
+
+  var additionalLinksFetchResponse = await fetch('http://localhost:4010/additional_links.json', {headers: headers});
+  // Handle error
+  if (!additionalLinksFetchResponse.ok) {
+    var message = `Error ${additionalLinksFetchResponse.status}: File not found or file not in correct format`;
+    alert(message);
+    throw new Error(message);
```
birm, line 575: I think that should the file not exist, the expected behavior should be equivalent to just adding no additional links.

birm (changes requested):

apps/viewer/init.js
Vedant1202: Alright I'll do that

Vedant1202: @birm Should I remove the alert then? And just keep it as throwing an error?

birm: More important is the idea that a bad/missing file should not stop the code from executing.

Vedant1202: Yes, I've done that. Also removed the alerts

apps/viewer/init.js
Vedant1202: Fixed that

birm (approved):
Thanks for working through this!

Issue #458, opened 2020-11-30
Upload ".jpg" images for visualization


Brayan-Valenzuela:
The data set I have contains .jpeg images, however when I upload them to the viewer, it says that this format is not compatible. 

As a solution, I thought of converting the .jpeg images to .tiff, since this format does accept it. However, you can see in the image the error that comes out when you try to upload it. 

![WhatsApp Image 2020-11-30 at 10 38 09 AM](https://user-images.githubusercontent.com/44003588/100630139-2ccf6680-32f8-11eb-8dbd-e81863591239.jpeg)

birm:
Thanks for the report.
It's unfortunately not enough for it to be just a tiff, it needs to be a pyramidal tiff. We've added [a route](https://github.com/camicroscope/SlideLoader/blob/62b4b0fba92c601e8ed95f52c49cd17020ff05fb/SlideServer.py#L65) that I don't think we use yet for this kind of conversion. Alternatively, [this link](https://libvips.github.io/libvips/API/current/Making-image-pyramids.md.html) may be useful. Please let me know whether or not this works for you.

Brayan-Valenzuela:
The solution was a perfect fit for our use case, thank you very much.

PR #512, opened 2021-04-02
Now, JPG images can be uploaded using batch loader

BeeramVineethReddy:
## Summary

In the previous version of the software, the user is not allowed to upload the images of jpg, jpeg, or png format.

But in many cases, the images to be visualized might be in the jpg format. It is a painful job to convert all the jpg image data to pyramidal tiff format, then upload and visualize. It would be easy for the user if we allow uploading of jpg images also.

## Motivation

the issue(regarding the uploading of JPG images) is reported as issue number: #458

### DIFF
--- a/apps/batchloader/batchLoader.js
+++ b/apps/batchloader/batchLoader.js
@@ -295,7 +295,7 @@ function updateFileName(oldfileName) {
     let newFileName = $('#newFileName');
     let newName = newFileName.val();
     let fileExtension = newName.toLowerCase().split('.').reverse()[0];
-    alert(fileExtension);
+  
     if (newName!='') {
       if (fileNames.includes(newName) || existingFiles.includes(newName)) {
         newFileName.addClass('is-invalid');
--- a/apps/batchloader/batchLoader.js
+++ b/apps/batchloader/batchLoader.js
@@ -11,7 +11,7 @@ let finishUrl = '../../loader/upload/finish/';
 let checkUrl = '../../loader/data/one/';
 let chunkSize = 5*1024*1024;
 let finishUploadSuccess = false;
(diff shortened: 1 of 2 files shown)


apps/batchloader/batchLoader.js
```diff
@@ -293,7 +295,7 @@ function updateFileName(oldfileName) {
     let newFileName = $('#newFileName');
     let newName = newFileName.val();
     let fileExtension = newName.toLowerCase().split('.').reverse()[0];
-
+    alert(fileExtension);
```
birm, line 298: Did you indend this just for dev purposes?

BeeramVineethReddy: Yes, I used it(line 298) just for testing purposes..!

BeeramVineethReddy: We can just comment it out if we don't want to see an alter popup in the browser while uploading.

birm:
```suggestion
```

akhil-rana:
Haven't been in regular touch with the development recently. So are JPEGs now supported by the server too?

apps/batchloader/batchLoader.js
BeeramVineethReddy: The code is modified as per the suggestions made .. with the commit message "Suggested Changes Made" ..!

birm (changes requested):
So, we've added some support for a quick conversion of jpg/png images primarily so that developers/new users don't need to find WSI images. However, this was added to loader.js in this commit - https://github.com/camicroscope/caMicroscope/commit/0554794ea063c5390da3fbd81223aa83e439bd41

However, this has not been carried over to the batch loader. For some reason I was under the impression that they were using the same code, but that seems not to be the case. As a result, this doesn't seem to work now.

For next steps, I'd see this as either reducing code duplication by using methods from loader.js for both loaders, or duplicating the functionality into batchLoader.js.

PR #527, opened 2021-06-13
GSoC Project Cross Slide Coordinated Viewing First Milestone

Hemansh31:
## Summary

In this milestone I have implemented the options for choosing which slide to view in the viewer page and have implemented pagination, sorting and filtering as was discussed with @nanli-emory. In the cross viewer I have implemented independent panning, zooming, adding labels, adding rulers, adding annotations , viewing layer managers for both the slides. using magnifier tool and the screenshot feature which takes screenshot of both the slides and joins them together. 

## Motivation

This is the first Milestone of my GSoC Project Cross Slide Coordinated Viewing.

## Testing

I have tested this feature in Google Chrome and Firefox and everything seems to be working fine.

PR #530, opened 2021-06-27
Second Milestone : Cross Slide Coordinated viewing

Hemansh31:
## Summary

Basic slide sync feature added. Adding rulers to screenshot feature extended to crossviewer. Rotation support added while adding rulers. All new js files are lint fixed.
## Motivation

This is in continuation to my previous PR #527
## Testing

I tested this on my local development setup

### DIFF
--- a/apps/crossviewer/dataloaders.js
+++ b/apps/crossviewer/dataloaders.js
@@ -1,166 +1,166 @@
 function layersLoader() {
-    // human
-    function loadingHumanOverlayers() {
-      $D.humanlayers = {};
+  // human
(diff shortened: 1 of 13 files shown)


birm:
Quick question, when you say rotation support, what do you mean?

Hemansh31:
> Quick question, when you say rotation support, what do you mean?

Earlier when the slide was rotated and then if we tried to add a ruler there was a buggy behaviour, that's what I have resolved!!!!

birm (approved):

PR #537, opened 2021-07-31
Include segmentation as annotations

r7rohan:
Segmentation extension
### Contains
- Converted Segmentation results to annotations using contours and saved it for future use.
- Fixed slide renaming (slidechangemodal wasn't working earlier)
- Hopefully fixed the bug in moving-points (multi-mode) and also fixed canvas bugs in segmentation.
- Trivial UI changes

ezgif com-gif-maker (2)

### Why?
Saving segmentation as annotations could be useful, as it can be used in future, and also could circle interesting areas as labels. Could also save whole image instead of contours as overlays but intentionally left it.  

### How?
I have added segmentation overlays which could be used both in viewer, side-by-side viewer and segmentation tabs. Currently, it can store name, notes and deletion can be done using popup-panel.

### DIFF
--- a/apps/segment/segment.html
+++ b/apps/segment/segment.html
@@ -135,6 +135,7 @@
   <script type='text/javascript' src='../../core/Store.js'></script>
   <script type='text/javascript' src='../../core/CaMic.js'></script>
 
+  <script type='text/javascript' src='../../core/extension/segment-annotation.js'></script>
   <script type='text/javascript' src='../../core/extension/openseadragon-canvas-draw-overlay.js'></script>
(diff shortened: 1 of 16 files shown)


birm (approved):
nice work and a natural addition to this workflow!

birm:
@r7rohan I didn't test this enough, and it ended up breaking support for existing segmentations, so I had to revert.

It's ok to just save these as human annotations, that way you don't have to rewrite too much logic hopefully. This PR probably shouldn't change viewer behavior at all.

Apologies for not catching this earlier.

PR #545, opened 2021-08-19
For 3.9.7

birm:
* Save Segmentations [#542](https://github.com/camicroscope/caMicroscope/pull/542)


birm:
postponed to focus on bugfix

PR #549, opened 2021-09-20
fix ui: added margin between dropdown and search bar

4rr0w:
Fixed UI on table.html page, added margin between search bar and dropdown menu

## Summary
before: 
Screenshot_2021-09-20 CaMicroscope Data Table
after:
Screenshot_2021-09-20 CaMicroscope Data Table(1)

## Motivation
While doing code refactoring, I noticed that there was no margin between the mentioned two components. 

## Testing
This does not affect any functionality.

## Questions
I was thinking about doing more such changes to improve the UI. If that's okay, please assign me the task.

### DIFF
--- a/apps/table.html
+++ b/apps/table.html
@@ -110,7 +110,7 @@
                     </div>
                     <div class="row" id="filters-check" >
                     </div>
-                    <div class="search-box float-left form-group d-md-inline-flex mb-4">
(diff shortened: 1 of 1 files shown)


birm (approved):
Thank you!

nanli-emory (commented):
😄

4rr0w:
@birm Can I use CSS to style components(if required) rather than Bootstrap. If yes, Can you please provide any reference I should follow.

Issue #546, opened 2021-09-04
nanoBorb Documentation Missing  


thesloppyguy:
**Is your feature request related to a problem? Please describe.**
In the README documentation there is a mention of nanoBorb, as a standalone application for individual users without a server. It has a linked attached to it for a section called "#nanoborb". As the section doesn't exist in the Document, so no redirects occur for the user. Leading to a dead end.

**Describe the solution you'd like**
If the documentation already exists, we should modify the link so it points to the correct documentation.
Otherwise if documentation doesn't exist, we should create one ASAP and mention the progress in the READ ME.  

**Describe alternatives you've considered**
If the documentation and the feature both don't exist, we should remove this portion "nanoBorb covers nanoBorb, the version of caMicroscope designed as a standalone application for individual users without a server." from README  to avoid user confusion in the future. 

**PS**
I would be happy to work on this issue. Guidance along the way would be appreciated.

Ref: https://github.com/camicroscope/caMicroscope/blob/master/README.md

manikmmalhotra:
@birm can i start with this issue if no-one working on this?

birm:
I'll always welcome PRs :)

chezhia:
Any updates on this issue? We are also interested in using this in standalone mode within our development container.

birm:
Hi @chezhia! We haven't rebuilt nanoborb in a while, but you can find it at https://github.com/sbu-bmi/nanoborb

PR #556, opened 2022-01-19
addressing issue #546  nanoBorb Documentation missing

A7-4real:
the added link that leads to nanoBorb Documentation

## Summary

A subtle change in Readme.md Introduction section which leads to nanoBorb documentation.
nanoBorb Doc issue

## Motivation

fixes open issue #546 

## Questions

1) Whether we should write documentation of nanoBorb in CaMicroscope's Readme.md file or its fine to redirect?
2) I'm trying to open this link to a new tab but it opens in the same tab even after using the HTML anchor target attribute.


birm:
I'm still thinking, but it may make more sense to remove the reference to nanoborb entirely. When we removed the section, I think we just missed it.

A7-4real:
I guess removing the reference could be a better option because User guide and Hosted setup are self-describing enough.

birm (approved):

PR #576, opened 2023-02-26
I replace all the var variables in to let variables in all javascript files

Karanjot786:


birm:
How did you test this?

birm:
let is not identical to var. If you're going to make a PR like this, you should run the application against your proposed fix and check all functionality.
Also, you've changed repositories, such as jquery. If you do want to make those changes, you should make them upstream first.
I also don't think this is worth doing for now. This kind of thing should be part of a larger refactor effort.

PR #601, opened 2023-03-15
Styled export.html 

Sankalp-sa:
## Summary
Just Styled the app/port/export.html

## Motivation
Before when we were going to export.html it felt like we were not anymore in the caMicroscope web application we came somewhere else. So, I just gave the color code and style which the website follows. 

##Before :
image

##After :
image

## Extra message :
This, is my first pull request to caMicroscope @birm so, please accept it 🥺 and let me know if somethings more to change in this issue.


birm (approved):
Looks reasonable to me!

PR #631, opened 2023-03-29
Update signup.html

Lokeshsri11:
can you review the new signup UI and code
https://loki-signup.vercel.app/?#


birm:
I appreciate you trying to get into open source, but this is not a good pull request. It contains lorem ipsum, and does not function as a replacement for the page you're proposing overwriting, if at all. Also, this seems to be from some kind of a template? If adapted, is it licensed freely enough for us to just use anyway?

Going forward, I'd recommend taking time to understand the code you're trying to change, and writing a PR with enough context in the comments to motivate the change, as well as describe what you've done.

PR #651, opened 2023-08-19
Naming only: refer to iipsrv instead of openslide to prevent confusion

CGDogan:
This code will also be used for bioformats


birm (approved):
I think this is fine, even though I think it's called openSlide because it opens a slide, and the fact that we use openSlide for iip is just a coincidence  (https://github.com/camicroscope/caMicroscope/commit/e5208068b8885c4651b0fff1c561451b328618c1)

In any case, this does resolve ambiguity, so approved.

PR #653, opened 2023-08-27
Bioformats extensions

CGDogan:


birm (approved):
Lgtm. Maybe in the future the loader should report which extensions it understands via api?

CGDogan:
Hmm, I think this would be wasteful in terms of bandwidth and also considering that these extensions change very rarely?

birm:
Fair enough!

PR #655, opened 2023-09-06
images in subdirs

CGDogan:
Corresponding to https://github.com/camicroscope/SlideLoader/pull/82

### DIFF
--- a/apps/batchloader/batchLoader.js
+++ b/apps/batchloader/batchLoader.js
@@ -21,8 +21,12 @@ $(document).ready(function() {
   store.findSlide().then((response) => {
     for (i=0; i<response.length; i++) {
       existingSlides.push(response[i].name);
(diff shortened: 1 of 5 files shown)


birm:
How should I use this to actually load a dicom image from the ui? Is that pattern batch loader specific?

CGDogan:
It calculates a hash from every DICOM file uploaded, a hash of the image identifier and not of the file. Hence if you would like to upload two files making up an image, it doesn't matter how you upload them. You can use the single file uploader and this would work even if you uploaded them with a 10 year delay between them.

Nevertheless still we'll have one entry in the list for each file uploaded even if they make up one image and it doesn't matter which one you click on to show the same image.

birm (approved):
looks good; while we'd prefer to have one database entry per slide, we'll need to decide how to do that.

Thank you!

PR #716, opened 2024-03-10
add gap between Signup and Feedback buttons

yzhong32:
## Summary

Add a gap between Signup and Feedback buttons.
From
Screenshot 2024-03-09 at 11 20 16 PM

To
Screenshot 2024-03-09 at 11 55 03 PM

## Motivation

Refine the UI to ensure elegance and consistency with the style of existing buttons.
## Testing

Yes. Tests passed.


birm (approved):
clean solution, nice!

PR #742, opened 2024-03-11
Improved error message visibility in the segment model upload interface

yzhong32:
## Summary
Improved error message visibility in the segment model upload interface by changing the font color from yellow to red for enhanced clarity and user experience.

Screenshot 2024-03-11 at 2 11 41 PM\
Screenshot 2024-03-11 at 2 03 35 PM

## Motivation

This change addresses the difficulty of reading error messages due to insufficient contrast with the background color.
## Questions

I've updated the error message color to red to increase visibility, although I'm open to feedback if this may not be the best choice.

### DIFF
--- a/apps/segment/segment.css
+++ b/apps/segment/segment.css
@@ -28,7 +28,7 @@ input:focus:invalid {
 }
 
 .error{
-    color: yellow;
+    color: red;
 }
 
 .switch {


birm (approved):
Simple, low stakes change with a clear benefit. Good job finding this!

Issue #757, opened 2024-03-13
🐛 Help button on workbench not displaying the user guide 


hannyP:
**Describe the bug**
when you click on the help button it should display the user guide but it didn't showing the following error
![Screenshot 2024-03-13 152336](https://github.com/camicroscope/caMicroscope/assets/141145167/e4b0d990-f84f-4b8b-a407-2cd507421484)

**To Reproduce**
1. Go to  workbench
2. Click on option and click on help
3. See error

**Expected behavior**
should display the readme guide 

**Screenshots**
![Screenshot 2024-03-13 152309](https://github.com/camicroscope/caMicroscope/assets/141145167/dbc97b73-c764-45bd-9eb1-1daf4849ec5a)

**Desktop (please complete the following information):**
 - OS: windows 11
 - Browser edge

Akhilesh-max:
It's been fixed and there's a PR open for the same 
 #683

PR #683, opened 2024-03-05
Fixes #606 : This Pull Request resolves the bug that was preventing the ML kit help-guide from functioning properly.

Akhilesh-max:
This Pull Request addresses the issue that was causing the ML kit help-guide to malfunction. The root cause was an improper usage of the ‘marked’ library, which is responsible for parsing Markdown into HTML.

Changed files :: `apps/dev-workbench/workbench.js` and `apps/dev-workbench/workbench.html`.

Before changes ::

https://github.com/camicroscope/caMicroscope/assets/124369508/f1a304a1-0860-4b3f-9f3e-5ec7d10f773f

After changes ::

https://github.com/camicroscope/caMicroscope/assets/124369508/6ea40b47-8787-4131-9172-ee00a213f915

PR #759, opened 2024-03-13
Fixed bug from issue  #757 , added showdown library and converted markdown to html in workbench.js

hannyP:
### fixed the bug

 I am an outreachy 2024 applicant and found this issue while running the project on my local machine  

## Testing
it hasn't been tested as only UI elements were being tested in the repo.

## Questions
I used the showdown library to solve this issue , let me know if i shouldn't have used any library do tackle this issue

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
--- a/apps/dev-workbench/workbench.html
+++ b/apps/dev-workbench/workbench.html
@@ -85,8 +85,7 @@
       
       <ul class="navbar-nav mr-auto px-2">
 =======
-      <!-- Navbar brand -->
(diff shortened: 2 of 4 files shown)


hannyP:
now when you click on help in workbench tab it should display the user guide like this
Screenshot 2024-03-13 164941

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
birm, line 1: Please don't include setting/config files for editors in PRs.

birm (changes requested):
Just one change and I think It'll look good!

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
...
+    "liveServer.settings.port": 5501
+}
```
birm, line 3:
```suggestion
```

birm:
turns out to be fixed with #683

Issue #760, opened 2024-03-13
Non-updated year in footer of apps/table.html


shusonialumnwi:
**Describe the bug**
They displayed year in the footer is a non-dynamic 2021, with this updated year, it suggests that camicroscope has not been updated since 2021 or is not being maintained. An expected behavior would have been for the year to reflect the current year.

**To Reproduce**
1. Go to 'http://localhost:4010/apps/table.html'
3. Scroll down to 'the end of the page'
4. See that the year displayed on the footer is 2021

**Expected behavior**
Rather than 2021, the copyright text in the footer should display the current year.

**Screenshots**
![Screenshot from 2024-03-13 12-39-38](https://github.com/camicroscope/caMicroscope/assets/162718653/3f6c0e36-b337-4ef7-8381-fceacd107ea8)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser chrome

arghya-align-pix-design:
i have solved this and put a pull request.

PR #761, opened 2024-03-13
modified date in footer to reflect the current year

shusonialumnwi:
* Created common/footer.js
* Removed static text and added a placeholder for copyright text in apps/signup/signup.html, apps/Info.html, and apps/table.html

## Summary

* Replaced the <p> tag containing the date with place holders with unique ids
* Wrote JQuery script (footer.js) to populate the placeholder with the current year

## Motivation

This is an attempt to ensure that dates in all  footers are consistent with the current year. #760

### DIFF
--- a/apps/Info.html
+++ b/apps/Info.html
@@ -167,9 +167,10 @@
 </div>
 
 <footer class="text-center text-white bg-dark p-3 mt-2">
-	<p class="p">Copyright © 2021 caMicroscope</p>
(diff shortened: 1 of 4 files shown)


birm (approved):
I think this is reasonable. I'm always a little nervous about dynamically populated copyright dates, but since these are on interactive pages, I think it's fine. Thank you. @shusonialumnwi

Issue #724, opened 2024-03-10
Login page


gbengaoluwadahunsi:
Summary:
This proposal aims to enhance the user experience on the caMicroscope platform by redirecting users to the login page when they click on the "Already have an account?" link on the Sign-Up HTML page.

Background:
Currently, when users click on the "Already have an account?" link on the Sign Up HTML page, they are directed to an empty or non-existent page. This can be confusing for users and may lead to frustration, as they expect to be redirected to the login page instead. By implementing this feature, we can improve user navigation and provide a more seamless experience for users transitioning between the Sign-Up and Login pages.

Use Cases:
New users who accidentally click on the "Already have an account?" link can be easily redirected to the login page without confusion.
Users who have previously signed up for an account but need to log in again can quickly access the login page from the Sign Up page. Improved user experience leads to higher user satisfaction and retention, contributing to the overall success of the caMicroscope platform.

Implementation:
To implement this feature, we need to update the HTML code of the Sign Up page to ensure that the "Already have an account?" link points to the correct login page URL. This can be achieved by modifying the anchor tag (<a>) associated with the link to direct users to the login page. 

Expected Impact:
The implementation of this feature is expected to have a positive impact on both users and the caMicroscope platform. By providing users with a clear and intuitive navigation path from the sign-up page to the Login page, we can improve user satisfaction and reduce frustration. This, in turn, can lead to increased user engagement and retention on the platform, ultimately benefiting the caMicroscope community as a whole.

Additional Notes:
It's important to ensure that the login page (e.g., login.html) exists and contains relevant content for users to log in successfully. Additionally, any changes to the HTML code should be thoroughly tested to ensure proper functionality across different devices and browsers. Regular user feedback and testing can help identify any issues or areas for improvement in the implementation of this feature.

shailpujan88:
Hey, I'm interested in tackling this issue. Could you please assign it to me? Thanks!

gbengaoluwadahunsi:
Kindly go through the PR to ensure those issues are resolved already.

birm:
Closing due to lack of activity.

PR #775, opened 2024-03-15
Fixed issue Login page #724

Ayesh-07:
Fixed issue #724 Login page 

## Summary
This proposal aims to enhance the user experience on the caMicroscope platform by redirecting users to the login page when they click on the "Already have an account?" link on the Sign-Up HTML page.

## Motivation
My motivation for this Pull Request stems from my desire to contribute to the caMicroscope organization as an Outreachy applicant. I am eager to enhance my skills and knowledge while making a meaningful impact. By fixing issue #724, I aim to improve the user experience on the caMicroscope platform. This enhancement redirects users to the login page when they click on the "Already have an account?" link on the Sign-Up HTML page, streamlining the user journey and making the platform more user-friendly.

signup

### DIFF
--- a/apps/signup/signup.html
+++ b/apps/signup/signup.html
@@ -109,7 +109,7 @@
           If you are an Admin, you can directly signup users.
         </p>
       </form>
-      <div class="text-center" onclick="loginPage()">Already have an account? </div>
+      <div class="text-center" onclick="loginPage()"><a href="/login.html">Already have an account?</a> </div>
     </div>
     <footer class="text-center text-white bg-dark p-3">
       <p class="p">Copyright © 2021 caMicroscope</p>
--- a/apps/signup/signup.html
+++ b/apps/signup/signup.html
@@ -109,7 +109,7 @@
           If you are an Admin, you can directly signup users.
         </p>
       </form>
-      <div class="text-center"><a href="/login.html">Already have an account?</a> </div>
+      <div class="text-center"><a href="../..//login.html">Already have an account?</a> </div>
     </div>
     <footer class="text-center text-white bg-dark p-3">
       <p class="p">Copyright © 2021 caMicroscope</p>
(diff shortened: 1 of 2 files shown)


birm:
This _should_ work via https://github.com/camicroscope/caMicroscope/blob/55d8db226ad28daa9908d00142fbbae4c61812e0/apps/signup/signup.js#L116

Does this not work in your experience?
Notably, it's likely you're developing in a way where there is no login and no login page, so it may go to a broken link, but it should try.

I'm not opposed to changing it to a regular link, but that'd involve cleaning up that loginpage function in addition to adding the href.

Ayesh-07:
@brim, there is no login page in the current setup. I attempted to update the JavaScript function, but the result doesn't look better. I've cleaned up the JavaScript code and am now waiting for your review.

apps/signup/signup.html
```diff
@@ -109,7 +109,7 @@ <h2 style="margin-top: -.5em;">User Signup</h2>
           If you are an Admin, you can directly signup users.
         </p>
       </form>
-      <div class="text-center" onclick="loginPage()">Already have an account? </div>
+      <div class="text-center"><a href="/login.html">Already have an account?</a> </div>
```
birm, line 112: I'd prefer to use relative paths whenever possible, as we often see deployments at urls like http://example.com/pathology/caMicroscope/

```suggestion
      <div class="text-center"><a href="../..//login.html">Already have an account?</a> </div>
```

birm (changes requested):
Just one edit!

apps/signup/signup.html
Ayesh-07: @birm I have made the requested change to add `../..//login.html`. Could you please review it and confirm if it aligns with your project's standards?

ujjwaldubey1:
@brim we should make a login page in the current setup so that we can provide that file path in the javascript function . I think most feasible solution from here is this.

PR #797, opened 2024-03-18
Created a Feedback Page

sarahannie:
Created a feedback page instead of using Google form 

## Summary

I have created a feedback page that is responsive to any screen size. This feedback page serves as the frontend view of what the feedback will look like, and it is divided into two sections: general feedback and bug reports.
Screenshot (198)
Screenshot (199)
Screenshot (200)
Screenshot (201)
Screenshot (202)

## Motivation

I have developed this feedback page, and upon approval, I aim to create the admin interface and establish the connection between the frontend and backend. This will allow the admin to conveniently view feedback directly from the website, eliminating the need to check Google Forms for feedback. This approach mitigates the risk of lost information, improves accessibility, and prevents clients from being discouraged from providing feedback due to navigating between different platforms.

## Testing

Yes, I have tested it with different screen sizes for easy compatibility

## Questions

No.

it ready for review


birm:
I don't quite understand, where does this feedback go?

sarahannie:
> I don't quite understand, where does this feedback go?

I want it to go to the admin dashboard by creating a page admin can receive the feedback directly. here is the issue I created for this https://github.com/camicroscope/caMicroscope/issues/853#issue-2203760139

PR #852, opened 2024-03-23
Fix : Redundant Cancel/Close buttons on modals used in table.html

harshpreet14:
## Summary
This PR simplifies the user interface across multiple modals within the application by removing redundant Close or Cancel buttons. Specifically, changes were made to the Delete Modal, Slide Name Change Modal, Update Slide Modal, and DICOM Server Modal. This decision enhances UI clarity and improves the overall user experience.

Screenshot 2024-03-23 114130
Screenshot 2024-03-23 114103

## Motivation

Having both Close (×) and Cancel buttons within the same modal can confuse users and detract from a streamlined interface design. By removing one of these elements, we aim to make the user's choices clearer and the UI more intuitive.

## Testing

Beyond the automated testing hooks, manual testing was conducted across all affected modals to ensure that:

- The remaining button (either Close or Cancel) functions correctly to dismiss the modal.
- User experience is not negatively impacted by the change.
- The UI looks cleaner and more consistent across the application.

## Questions

 While this change is aimed at improving the UI, there's always room for feedback and further discussion:

- Are there any specific modal dialogs where the presence of both buttons might still be justified?
- How does this change impact accessibility, and are there any additional adjustments needed to maintain or improve accessibility standards?

### DIFF
--- a/apps/table.html
+++ b/apps/table.html
@@ -171,9 +171,6 @@
                 <div class="modal-content">
                     <div class="modal-header">
                         <h5 class="modal-title" id="title-of-dialog-dicom">DICOM Server</h5>
-                        <button type="button" class="close" onclick="initialize();" data-bs-dismiss="modal" aria-label="Close">
-                            <span aria-hidden="true">&times;</span>
-                        </button>
                     </div>
                     <div class="modal-body">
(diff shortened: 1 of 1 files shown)


birm:
This is a good idea, but I kind of like the flexibility by having close options in the three ways users may expect (like a window, as an option, escape key)

PR #868, opened 2024-03-25
Adding a loader

SanjanaSogimatt:
Adding a loading animation on window load
## Summary

Built with html nad pure css
## Motivation

Whenever it took time to load it always showed a blank white screen hence I added loader to it

https://github.com/camicroscope/caMicroscope/assets/87171143/2584daad-9a62-4427-85c5-c9a8dab73c49

### DIFF
--- a/apps/landing/landing.html
+++ b/apps/landing/landing.html
@@ -12,7 +12,13 @@
 
 </head>
 <body>
-
+<!-- loading animation -->
+<div class="main-site">
+    <!-- adding css -->
+    <div class="main-content" role="main" id="mainContent">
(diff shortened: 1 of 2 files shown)


birm:
This is pretty cool, but I'm curious in your experience about how long this loader is on screen (especially versus the white screen before).

SanjanaSogimatt:
Well, I loader will be displayed until the window is fully loaded. It has a smooth transition after the window is loaded completely

birm:
I'll merge this, but I'm still curious how long it lasts for you, especially since for me I don't see the loader for more than what seems like a frame or two.

PR #903, opened 2024-03-28
Added Aria Labels to Image Links

### LINKED ISSUE #901: [Accessibility] Links don't have discernible text
Some image links on the Home Page don't have discernible text. This is not following the [WCAG 2.1 Standard](https://www.w3.org/TR/WCAG21/#link-purpose-in-context) regarding links. These links are inaccessible to these users, especially those who rely exclusively on a keyboard (and no mouse) to navigate a webpage.

**To Reproduce**
1. Right-click on the home page and click inspect to open the dev tools window
3. If you have the Axe extension installed, open it and run an accessibility test
4. One of the accessibility issues discovered is that "links must have discernible text"

**Expected behavior**
According to the WCAG 2.1 standard, link text and alternate text for images, when used as links, must be discernible by a screen reader, must not have a duplicate label, and must be focusable.

**Screenshots**
links discernible text

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Microsoft Edge

nwanduka:
I added aria labels to the images (which also happen to be links) on the home page. Those images didn't initially have aria labels which posed an accessibility issue for users who use only keyboards to navigate the site.

## Motivation
This PR fixed #901

### DIFF
--- a/apps/landing/landing.html
+++ b/apps/landing/landing.html
@@ -81,13 +81,13 @@
 
         <div class="content">
             <p>caMicroscope is a tool to view, annotate, and analyze biomedical images.</p>
-            <a href="#" class="image fit"><img src="./banner1.jpg" alt="" /></a>
+            <a href="#" class="image fit" aria-label="Reload Home Page"><img src="./banner1.jpg" alt="" /></a>
 
         </div>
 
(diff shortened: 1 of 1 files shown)


birm (approved):
good catch!

PR #856, opened 2024-03-23
Fixed issue #837  HTML element does not have lang attribute 

Ayesh-07:
Fixed issue #837 

## Summary

HTML element does not have Lang attribute #837 
To fix this issue, I added the `lang` attribute to the `html` element and specify the appropriate language code that reflects the primary language of the content on the page. For example, for English content, I would use `lang="en"` in the `html` tag.

## Motivation

I'm passionate about contributing to the caMicroscope organization to further my skills and knowledge. One of the issues I've successfully fixed is [Accessibility] HTML element does not have lang attribute #837. I would appreciate a review of my work to ensure it meets the project's standards. This experience has deepened my understanding of user-centered design and my ability to create intuitive user interfaces.

PR #915, opened 2024-03-29
Added a language attribute

gbengaoluwadahunsi:
Summary
The language attribute has been added to the HTML page to specify the language of its content. This helps browsers and search engines understand the language used on the page and improves accessibility for users.

Motivation
The motivation behind adding the language attribute is to provide metadata about the language used in the HTML page, enhancing its semantic structure and accessibility. This ensures that users with different language preferences or assistive technologies can properly interpret the content.

Testing
The addition of the language attribute has been tested to ensure it functions correctly across different browsers and devices. Additionally, it has been validated using HTML validation tools to confirm compliance with web standards and best practices.

### DIFF
--- a/apps/Info.html
+++ b/apps/Info.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 
 <head>
 	<meta name="keywords" content="camicroscope, quip" />


birm:
This was done in #856 as merged into develop! Thanks for submitting anyway!

PR #933, opened 2024-03-31
Modifying toolbar title

SanjanaSogimatt:
The pop up on the toolbar was not efficient and it blocked other components while displaying the title of the tool. Hence modified it, making it more user friendly
## Summary
Below is the changes, I changed the background color to dark as it would be easier to read and made sure that the title is not overlapping with other components and removed the default title pop up.

### **Before:**
https://github.com/camicroscope/caMicroscope/assets/87171143/23c577b2-31f1-4c2b-8e2a-d47a270afa41

### **After:**

https://github.com/camicroscope/caMicroscope/assets/87171143/41add129-f3c0-4e7a-8e1b-207919a3337b

## Questions

If there are any questions or suggestions do feel free to comment down below

### DIFF
--- a/components/toolbar/toolbar.css
+++ b/components/toolbar/toolbar.css
@@ -71,7 +71,7 @@ ul.drop_down {
 }
 
 ul.tools > li:hover ul.drop_down {
-  flex-direction: column;
+  flex-direction: row; /* Consumes less space */
   display: flex;
   width: fit-content;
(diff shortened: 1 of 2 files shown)


birm (approved):
For posterity, I'll note that in the situations where there's a dropdown from the tool, it's not really better, but otherwise (for most tools) it is! Good suggestion and change!

PR #938, opened 2024-04-01
Modifying table page

SanjanaSogimatt:
I modified the table.html file for responsiveness on all different orientation
## Summary

Adding responsiveness to the table.html file to control overflow and making it user friendly
## Motivation

Motivation behind this is that when you open table.html page in different sizes of the screen it overflows the content and it is not user friendly.
## Questions

If there are any questions please feel free to drop in comments

### DIFF
--- a/apps/table.css
+++ b/apps/table.css
@@ -174,6 +174,7 @@ nav li:not(.active):hover{
 .bg-gray {
   background-color: #eee;
 }
+
 @media (max-width: 640px) {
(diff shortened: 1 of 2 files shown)


SanjanaSogimatt:
I am still working on it

birm:
It looks like since I took so long to review, this issue has since fixed in another PR. Thanks for the contribution in any case!

PR #957, opened 2024-04-12
Created GH workflow for accessibility checks

nwanduka:


birm:
Since caMicroscope is already in raw html/js form, we typically just use a http server to host it (see https://github.com/camicroscope/caMicroscope/blob/55d8db226ad28daa9908d00142fbbae4c61812e0/.github/workflows/node.js.yml#L30)

hope this helps!

nwanduka:
> Since caMicroscope is already in raw html/js form, we typically just use a http server to host it (see
> 
> https://github.com/camicroscope/caMicroscope/blob/55d8db226ad28daa9908d00142fbbae4c61812e0/.github/workflows/node.js.yml#L30
> 
> )
> hope this helps!

It does help a lot. Thank you so much for the tip. Now I'm past the start script error. There's a challenge though when it starts to run the axe core CLI. It can't find the server hosting the application. I'll try to debug this on my own but if the issue persists, I'll be sure to reach out again.

nwanduka:
After two (may have been more) weeks and 17 commits, I'm thrilled to announce that I've finally had a breakthrough with this task. The workflow works!!! 🤗

image


@birm for your review. I also want to confirm that I'm using the right port. I had been using localhost/4010 and kept encountering errors. So I used this:
https://github.com/camicroscope/caMicroscope/blob/55d8db226ad28daa9908d00142fbbae4c61812e0/test/config.js#L4

birm (approved):

birm:
This is great! We'll have some test failures until we get them resolved, but fantastic work!

PR #902, opened 2024-03-28
Saving Marks Should Include User's X, Y, Zoom, etc #844

YukaUU:
## Summary
Added X, Y (image coordinate ）and zoom to the pop-panel
## Motivation
Because until now, it was not possible to check the image coordinate and zooming information of the area annotated in mouse mode.

## Testing
Before
Before_X_Y_Z
After
After_X_Y_Z

## Questions
please check whether my PR matches your issue #844 or not.

Issue #844, opened 2024-03-22
Saving Marks Should Include User's X, Y, Zoom, etc


birm:
Currently we have a post to log every so often. We should include this information about how the user is annotating in the annotations (Mark) too.

birm:
solved with #902 :)

birm:
solved with #966

Issue #943, opened 2024-04-03
Mark Point Editing Should Only Be Active if actively activated


birm:
The points editing conflicts with other viewer functions, and it is easy to accidentally mess with annotations.

birm:
It also ghosts incorrectly if you change zoom
<img width="940" alt="Screenshot 2024-04-03 at 4 09 31 PM" src="https://github.com/camicroscope/caMicroscope/assets/4951803/5b1d11de-7ae7-4d13-828f-6900e1bd2dc8">

lbhnsh:
Hi, can you assign this issue to me?

PR #958, opened 2024-04-20
Revert "Revert "Saving Marks Should Include User's X, Y, Zoom, etc #844""

birm:
Reverts camicroscope/caMicroscope#942 which reverted #902 
Apparently we're indecisive here. Making this for deeper test/review


birm:
I'm merging this back in. The issue is that clicking the annotation triggers a update, re #943 ; so the issue is there and not here.

Apologies to @YukaUU who did the task correctly, but I got jumpy seeing the bug caused with this change interacting with #943 and reverted.

PR #980, opened 2024-06-19
added aria-label to "more" buttons

### LINKED ISSUE #979: [Accessibility] Home Page: More buttons don't provide context to screen readers
**Describe the bug**
The "more" buttons on the home page are both read as "more" by screen readers. This doesn't provide context to a user navigating the site with a screen reader as to the function of the button.

**Expected behavior**
The buttons should be appropriately labelled so that the user know what to expect when they click it. For instance, the button that takes users to the documentation page should read "go to documentation" or "visit documentation page" or any text that provides better context.

**Screenshots**
image

nwanduka:
This PR fixes #979


birm (approved):
🌟

PR #991, opened 2024-07-01
Added pa11y and lighthouse a11y testing to workflows

nwanduka:
## Summary

Created pa11y.yml and lighthouse.yml to complement the already existing axe a11y testing tool in workflows

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -1,0 +1,31 @@
+name: lighthouse
+on:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
+    branches:
+      - master
+      - develop
+jobs:
+  lighthouse:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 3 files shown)


birm:
Am I reading the lighthouse report correctly that it seems to be stuck at the redirect?

nwanduka:
> Am I reading the lighthouse report correctly that it seems to be stuck at the redirect?

Yep! That's the issue I meant when I said it wasn't testing all the pages. I wonder why that is.

birm:
It looks like it's not following links or anything of the sort. Do we need to give it a list of urls?

nwanduka:
No, I don't think we need to provide it with a list of URLs. According to [the docs](https://github.com/GoogleChrome/lighthouse-ci/blob/main/docs/getting-started.md#collect-lighthouse-results), Lighthouse will "automatically find your project's static assets, run Lighthouse 3 times on each HTML file, and upload the reports to temporary public storage where they'll be accessible to anyone with the URL."

Issue #1002, opened 2024-07-03
[Accessibility] [UX] Slide Viewer Page - Annotation Name Suggestion


nwanduka:
**Describe the bug**
When creating an annotation on the Slide Viewer page without giving it a name, the error message indicates that the field must have a value but does not offer a suggestion on the expected format of the value.

**To Reproduce**
1. Open a slide.
2. Select the draw tool
3. Leave the identity name field empty and save.

**Expected behavior**
An error message should appear with a suggestion indicating the expected form of the value (e.g., "Please provide a name for the annotation. The name should be descriptive and unique.")

**Screenshots**
![image](https://github.com/camicroscope/caMicroscope/assets/122698422/d5b85049-cb68-4dc3-9cb1-82a0b11fefaa)

PR #1006, opened 2024-07-08
Fixed the Name Suggestion error on the Slide Viewer Page

Joice-crypto:
## Summary
This PR resolves this  #1002
The part of the code where it says that the field must be a value has been changed, the change now makes it clearer to the user what the expected format is for that field

## After

Captura de tela de 2024-07-07 22-57-28


birm (approved):
Thank you, I do think this is more descriptive!

PR #1029, opened 2024-08-14
Fixed keyboard accessibility issue on workbench page

### LINKED ISSUE #1022: [Accessibility] Workbench: Back button unresponsive to 'enter' key
**Describe the bug**
The back button on the workbench page can receive keyboard focus when tabbing through the page. However, when the button has focus and the Enter key is pressed, it does not perform its expected action of returning to the previous page. This issue impacts keyboard accessibility, as users who rely on keyboard navigation cannot use the back button functionality.

**To Reproduce**
1. Navigate to the workbench page
2. Use the Tab key to move focus to the back button
3. Press the Enter key

**Expected behavior**
Pressing Enter should activate the back button, returning the user to the previous page.

**Suggested investigation**
Check the event listener for the Enter key press on the back button.
Verify that the button's onClick event (or equivalent) is properly set up for keyboard interactions.

**Additional context**
The focus visibility of the button has been fixed in a previous update. This issue specifically relates to the button's lack of response to the Enter key.

nwanduka:
## Summary

This PR fixes #1022

### DIFF
--- a/apps/dev-workbench/workbench.css
+++ b/apps/dev-workbench/workbench.css
@@ -67,9 +67,8 @@
     margin-left: 0 !important;
   }
 }
-.navbar .nav-item .dropdown .options-button {
-  background-color: #000;
-  color: #fff;
+.navbar .nav-item .options-button {
+  color: #000;
   border: none;
 }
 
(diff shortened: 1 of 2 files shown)


birm (approved):
good idea!

Issue #1034, opened 2024-08-20
[Accessibility] Export Page: Element has low color contrast


nwanduka:
**Describe the bug**
The small text underneath the "Export Results" heading doesn't have sufficient color contrast.

**To Reproduce**
1. Go to 'Export' in the nav bar
2. Run the Axe a11y test or Lighthouse
3. See error

**Expected behavior**
Elements should have sufficient color contrast.

**Screenshots**
![image](https://github.com/user-attachments/assets/3c106571-80f1-474f-82a9-02534aaf1881)

Priyankagarai:
Hi @nwanduka I hope you’re doing well. I’ve created a pull request for the low color contrast issue Could you please take some time to review it?
I would appreciate it if you could provide any feedback or suggestions. Once reviewed, could you also merge it if everything looks good?

PR #1035, opened 2024-08-25
Fixed sufficient color contrast

Priyankagarai:
Closes https://github.com/camicroscope/caMicroscope/issues/1034

Element is having low color contrast issue, I have made necessary modifications with font style to meet sufficient color contrast.

### DIFF
--- a/apps/port/export.html
+++ b/apps/port/export.html
@@ -9,6 +9,9 @@
     .export--header{
       background-color: #17a2b8 !important;
       color: white;
+      font-weight: bold;
+      font-size: 14pt;
+
     }
     #output{
       padding: 20px;


birm:
Hi and thank you for opening this PR!

Can you open a report or something stating that this fixes color contrast? It doesn't look to me like you're changing color at all.

A list of tools can be found here: https://www.w3.org/WAI/test-evaluate/tools/list/

birm (changes requested):
Hi and thank you for opening this PR!

Can you open a report or something stating that this fixes color contrast? It doesn't look to me like you're changing color at all.

A list of tools can be found here: https://www.w3.org/WAI/test-evaluate/tools/list/

Priyankagarai:
Hi, thank you for your feedback!

I’ve tested the page using Lighthouse in Chrome DevTools. Initially, Lighthouse flagged poor color contrast (as shown before the fix), but after correcting the colors, the issue no longer appears, meaning the contrast now meets the accessibility standards.
Since Lighthouse doesn’t display warnings once the issue is fixed, there’s no message in the updated report. 

Let me know if further clarification is needed!

birm (approved):

PR #1053, opened 2024-12-25
Fix #726 Update slide upload instructions and error message for missing file extensions

### LINKED ISSUE #726: [UX] Unclear Error Message
I encountered this error message (screenshot is attached) when uploading a slide. My error is that I didn't include the file extension in the file name field. The error message is unclear and potentially confusing for users. It lacks specificity and does not provide actionable information on how to resolve the issue.

image

Here’s what I suggest can be done to improve the error message:

1. We can point out in the "Steps to upload a file" (and by extension the contributing docs) which file formats are supported and how users can ensure compatibility. Or in the second line, we could give an example of a compatible file name. So the second line would read: Provide the destination filename with extension. For example "sample.svs"
image

2. We can provide instructions on what users can do to rectify the issue and move forward. A better error message could read: The file name you provided is incompatible. File names should follow this format "filename.ext"

chandrikarj:
Updated file upload instructions and improved error message for missing file extension

## Summary
This PR improves the error message displayed when a user attempts to upload a file without including the file extension. The original error message was unclear, hence, it is updated it to be more specific and helpful. The new error message offers clear instructions on how to resolve the issue, including an example of a correct file name format. Additionally, the "Steps to upload a file" section has been updated to guide users on how to properly name their files.

_Changes Made:_

Updated error message: The error message now reads:
"The file name you provided is incompatible. File names should follow this format: filename.ext"
This provides clarity on the correct file naming convention.

Updated "Steps to upload a file" section:
The second line of the guide has been updated to:
"Provide the destination filename with extension. For example, sample.svs."
This emphasizes the need for users to include the file extension when naming their files.

Before:
Before-CA

After:
After-CA

## Motivation
The error message for missing file extensions was insufficient and potentially confusing for users. This update ensures the error message is actionable and provides users with the information they need to proceed with uploading files correctly. It fixes the open issue https://github.com/camicroscope/caMicroscope/issues/726 , [UX] Unclear Error Message.

## Testing
The new error message was tested by attempting to upload a slide without a file extension. The message now appears as intended, providing users with the necessary information to rectify the issue. Additionally, the updated instructions in the "Steps to upload a file" section are now present while uploading.

## Questions
Are there any other areas in the upload process that could benefit from clearer instructions or error messages? I would appreciate any feedback on the changes made and suggestions for further improvements.

### DIFF
--- a/.github/workflows/stale_issues.yml
+++ b/.github/workflows/stale_issues.yml
@@ -1,0 +1,17 @@
+name: 'Stale Issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
+          stale-pr-message: 'This PR is stale because it has been open 45 days with no activity.'
+          close-issue-message: 'This issue was closed because it has been stalled for 5 days with no activity.'
+          days-before-stale: 30
+          days-before-close: 5
+          days-before-pr-close: -1
--- a/apps/table.html
+++ b/apps/table.html
@@ -212,7 +212,7 @@
                                 <h3>Steps for uploading.</h3>
                                 <ul>
                                     <li>Select the file you wish to upload. The <b>Token</b> field should be automatically populated .</li>
-                                    <li>Enter the desired filename, including its extension, for the destination file.</li> 
+                                    <li> Provide the destination filename with extension. For example "sample.svs"</li> 
                                     <li>Click on the <b><i class="fas fa-arrow-right"></i></b> button to preview the slide.</li> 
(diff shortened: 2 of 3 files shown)


birm:
I'm not seeing this button consistently when I test this, see below with similar results on other pages.

https://github.com/user-attachments/assets/2fc5e8a5-3e25-4316-a009-0fb096e99f40

birm (changes requested):
I'm not seeing the same results you are; I see the change to the top instrucitons, but have not noted anything else described.
Screenshot 2025-01-22 at 10 47 16 PM

chandrikarj:
Hi @birm 

Earlier, the error message would only show after pressing Enter when the filename had no extension. I've now updated it so that pressing Enter is no longer needed.  

Here’s what it does now:  
1. It checks if the extension is in the list of allowed extensions. If not, it immediately shows the error message: `"extension" + " files are not compatible"`.  
2. If the extension is valid but the filename is missing an extension, it now displays the error message: `"The file name you provided is incompatible. File names should follow this format: filename.ext"` as soon as the user moves to the next input field.

https://github.com/user-attachments/assets/d945c731-64d1-4a2a-99c4-97ff2997ea01

PR #1083, opened 2025-02-18
test a11y test as warn not fail

birm:
Will this work? If it does, I'll merge it :)


birm:
I think this is _ok_.  We may someday want to add a PR comment to the fail condition of the axe workflow since the warning is a little less visible than I'd like.

